PR #5, opened 2021-06-16
fix: don't push the mfe image to docker.io

regisb:
The openedx-mfe Docker image should be re-built by every user. If we push the
image to docker.io, people are at risk of actually downloading it, and it will
simply not work for them.

This is ready for review @overhangio/tutor-developers.


BbrSofiane (commented):
LGTM

PR #7, opened 2021-07-04
Added MFE service (patch nginx-extra)

insad:
Added MFE service in patch nginx-extra, so a reverse proxy has access to the micro frontends on {{ MFE_HOST }}

### DIFF
--- a/tutormfe/patches/nginx-extra
+++ b/tutormfe/patches/nginx-extra
@@ -1,0 +1,19 @@
+# MFE service
+upstream mfe-backend {
+    server mfe:8002 fail_timeout=0;
+}
+server {
+    listen 80;
+    server_name {{ MFE_HOST }};
+
+    # Disables server version feedback on pages and in headers
+    server_tokens off;
+
+    client_max_body_size 10m;
+
+    location / {
+    proxy_set_header Host $http_host;
+    proxy_redirect off;
+    proxy_pass http://mfe-backend;
+    }
+}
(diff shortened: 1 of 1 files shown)


tutormfe/patches/nginx-extra
```diff
@@ -0,0 +1,19 @@
...
+
+    # Disables server version feedback on pages and in headers
+    server_tokens off;
+
+    client_max_body_size 10m;
+
+    location / {
+    proxy_set_header Host $http_host;
+    proxy_redirect off;
+    proxy_pass http://mfe-backend;
```
regisb, line 17: Please fix indentation.

regisb (changes requested):
I think that we should switch the proxy in the main Caddyfile configuration, such as traffic is always redirected to nginx, and no longer to the mfe container, even when not running a proxy on the host.

To clarify, in the absence of web proxy on the host, traffic is transferred as follows:

    Caddy -> MFE container (Caddy service)

I think that it should become:

    Caddy -> Nginx container -> MFE container (Caddy service)

This approach would be more consistent with the situation when Caddy is disabled.

regisb:
@insad did you understand my last comment about adding a Caddy proxy configuration? If yes, do you have the time to take this on? If not I can do it.

insad:
Please can you do it, no time at the moment here. Besides I don't have a dev setup at hand for testing.

regisb:
Closed in favour of #8.

PR #9, opened 2021-09-06
docs: waffle flags need to be deleted after uninstall

regisb:
In particular, if the profile waffle flag is not deleted, it will result in 500
errors in the LMS.

This is ready for review @overhangio/tutor-developers.


BbrSofiane (approved):
Looks good to me!

PR #16, opened 2021-10-18
docs: add instructions to customise MFE logos

BbrSofiane:
Make Logo URLs configurable using config.yml.

See [thread](https://discuss.overhang.io/t/customise-mfes-branding/2029).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -112,6 +112,27 @@ This plugin makes it possible to change existing and add new translation strings
 
 Your custom translation strings should now appear in your app.
 
+Customise MFEs Logos
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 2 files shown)


BbrSofiane:
@regisb I've made the changes to the README.

About improving the default logo, did you have any suggestions? I considered using something similar to your implementation of LazyStaticAbsoluteUrl but that would require changes to the tutor core. Is my assumption correct?

regisb:
Damn! I keep forgetting that the logo url depends not just on the LMS_HOST, but also on the theme that is in use on the lms. 

I think it's a shame that we have to jump through so many hoops just to fetch a simple static asset. I wish there was a view in edx-platform that 301-redirected `/theme/images/logo.png` to the actual logo url. Maybe we can make such a change upstream? For instance in the "openedx/core/theming" app?

I'll open a PR upstream. In the meantime, I'll merge your change.

PR #21, opened 2021-10-28
fix: incorrect MFE language user preference

regisb:
MFEs were always displayed in English, whatever the user preferred language. To reproduce:

1. Open user account: https://apps.demo.openedx.overhang.io/account
2. Switch to Español (Latinoamérica)

The account MFE should automatically switch to Spanish, but it remained in
English. This was due to the fact that the language cookie name was incorrectly
set to "language-preference" instead of "openedx-language-preference", which is
the default value in edx-platform.

To fix this issue, we could have chosen to set the LANGUAGE_COOKIE_NAME setting
to "language-preference" in tutor core. But since this value was recently set
to "openedx-language-preference" in nightly, we decided to make a corresponding
change here.

See discussion: https://discuss.overhang.io/t/all-mfe-in-english-cant-either-change-language-in-account-settings-12-1-5/2039/


BbrSofiane (approved):
LGTM

PR #34, opened 2022-01-14
avoid using two different Caddy base images

iamCristYe:
Currently openedx itself and mfe are using two different Caddy base images, which is not ideal. This PR fixes this.


regisb (approved):

regisb:
Thanks for the PR!

PR #36, opened 2022-02-10
fix: Escape single quotes inside the platform name.

feanil:
Because this file is sourced as bash, if there are single quotes inside
the platform for apostrophes(eg "Feanil's Dev Setup") then the config
file becomes malformed.

Fix it by escaping any apostrophes in the platform name.


regisb (approved):

regisb:
Great, thanks for that!

kdmccormick:
I think we need to take one step further into quoting insanity: https://github.com/overhangio/tutor-mfe/pull/38

PR #41, opened 2022-03-13
indentation added to mfe plugin example

CodeWithEmad:
Using the example of `mfe_branding_plugin` caused the error below:
`Error: Invalid patches in plugin mfe-branding. Expected dict, got <class 'NoneType'>.`
it looks like you missed a couple of spaces before the patches.

it has been tested and it solve the issue.


regisb (approved):
Good catch!

PR #46, opened 2022-04-20
feat: deprecate `runserver` in favor of `start`

kdmccormick:
See individual commit messages.

Blocks https://github.com/overhangio/tutor/pull/644

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -192,19 +192,37 @@ MFE development
 
 Tutor makes it possible to run any MFE in development mode. For instance, to run the "profile" MFE::
 
-    tutor dev runserver profile
+    tutor dev start profile
 
 Then, access http://apps.local.overhang.io:1995/profile/u/YOURUSERNAME
 
-To run your own fork of an MFE, start by copying the MFE repo on the host::
+You can also bind-mount your own fork of an MFE. For example::
 
-    tutor dev bindmount profile /openedx/app
+    cd /path/to/frontend-app-profile
+    npm install  # Ensure NPM requirements are installed into your fork.
+    tutor dev start --mount=. profile
 
-Then, run a development server that bind-mounts the repo::
+The changes you make to your fork will be automatically picked up and hot-reloaded by your development server.
 
-    tutor dev runserver --volume=/openedx/app profile
+This works for custom MFEs, as well. For example, if you added your own MFE named frontend-app-myapp, then you can bind-mount it like so::
 
-The changes you make to ``$(tutor config printroot)/volumes/app/`` will be automatically picked up and hot-reloaded by your development server.
+    cd /path/to/frontend-app-myapp
+    npm install
+    tutor dev start --mount=. myapp
+
+However, if you try to bind-mount an unknown MFE, you will see a Docker Compose error such as::
+
+  ERROR: The Compose file is invalid because:
+  Service myapp has neither an image nor a build context specified. At least one must be provided.
+
+Please note that bind-mounting a fork is only available for development (``tutor dev ...``), since production MFEs are compiled and served out of a single container. If you want to use a fork of an MFE in production, then you will need to set the repository URL in ``$(tutor config printroot)/config.yml``::
+
+    MFE_PROFILE_MFE_APP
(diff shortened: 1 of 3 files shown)


tutormfe/plugin.py
```diff
@@ -58,6 +58,25 @@
...
+    it's available in the APPNAME service container. This is only applicable
+    in dev mode, because in production, all MFEs are built and hosted on the
+    singular 'mfe' service container.
+    """
+    prefix = "frontend-app-"
+    if name.startswith(prefix):
+        # Assumption:
+        # For each repo named frontend-app-APPNAME, there is an associated
+        # docker-compose service named APPNAME. If this assumption is broken,
+        # then Tutor will try to mount the repo in a service that doesn't exist.
```
kdmccormick, line 74: I am torn on this and would appreciate your input @regisb : 

Currently, I allow `--mount` to be specified for any arbitrary `frontend-app-` repo. This is nice, because it would allow a user who has configured a custom MFE (say, frontend-app-authn) to bind-mount it without any hassle. The downside, though, is that we get a fairly confusing error message if the custom MFE hasn't actually been configured, since the custom MFE's service would be missing from env/dev/docker-compose.yml:
```
$ tutor dev start -d -m ../frontend-app-learning/ -m ../frontend-app-authn/
Bind-mount: /home/kyle/openedx/frontend-app-learning -> /openedx/app in learning
Bind-mount: /home/kyle/openedx/frontend-app-authn -> /openedx/app in authn
docker-compose -f /home/kyle/.local/share/tutor/env/local/docker-compose.yml -f /home/kyle/.local/share/tutor/env/dev/docker-compose.yml -f /home/kyle/.local/share/tutor/env/local/docker-compose.tmp.yml --project-name tutor_dev up --remove-orphans --build -d
ERROR: The Compose file is invalid because:
Service authn has neither an image nor a build context specified. At least one must be provided.
```

Options I see:
* Keep it as-is.
* Keep it as-is, but try to improve the error message in Tutor.
* Change it to a hard-coded allowlist of bindmounts (for now: profile, account, learning, gradebook).
* Add a `config` argument to COMPOSE_MOUNTS, which would allow us to see which MFE have been configured.

README.rst
```diff
@@ -192,19 +192,17 @@ MFE development
...
 
 Then, access http://apps.local.overhang.io:1995/profile/u/YOURUSERNAME
 
-To run your own fork of an MFE, start by copying the MFE repo on the host::
+You can also bind-mount in your own fork of an MFE. For example::
 
-    tutor dev bindmount profile /openedx/app
+    cd /path/to/frontend-app-profile
+    npm install  # Ensure NPM requirements are installed into your fork.
+    tutor dev start profile --mount .
```
regisb, line 203: I believe this line should be:

    tutor dev start --mount=. profile

tutormfe/plugin.py
regisb: I think that your approach is sensible. It's true that the error message is difficult to understand, but we expect that users who attempt to bind-mount volumes are rather advanced. Maybe we can simply add a note to the README?

I'm more concerned by the fact that this bind-mount will not work in `tutor local`, because the container is then named "mfe" and not `app_name`. Maybe we should add another note about this in the README.

README.rst
kdmccormick: Click allows either syntax, but I'll change it just to be consistent with the Tutor's dev docs on mounting.

kdmccormick: Done

tutormfe/plugin.py
kdmccormick: Good points. Done.

kdmccormick:
@regisb This is ready for another look.

regisb (approved):
Somehow I forgot to review this PR again -- sorry about the delay!

regisb:
Can you please rebase your changes and use `tutor_hooks` instead of `hooks` @kdmccormick?

kdmccormick:
I rebased, and I'm trying to manually test this again, but I'm stuck on an unrelated error currently.

PR #51, opened 2022-05-12
fix: Add env variable to enable records button in profile

ghassanmas:
Learning-MFE won't view records button, unless the
  ENABLE_LEARNER_RECORD_MFE is set to true.
  This shall fixes: openedx/build-test-release-wg/issues/168

### DIFF
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -38,6 +38,11 @@ config = {
             "name": "profile",
             "repository": "https://github.com/edx/frontend-app-profile",
             "port": 1995,
+             "env": {
+                "production": {
+                    "ENABLE_LEARNER_RECORD_MFE": "true",
+                },
+            },
         },
     },
 }


tutormfe/templates/mfe/build/mfe/env/production
```diff
@@ -22,5 +22,6 @@ SEGMENT_KEY=''
 SITE_NAME='{{ PLATFORM_NAME|replace("'", "'\\''") }}'
 STUDIO_BASE_URL='{{ "https" if ENABLE_HTTPS else "http" }}://{{ CMS_HOST }}'
 USER_INFO_COOKIE_NAME='user-info'
+ENABLE_LEARNER_RECORD_MFE='true'
```
regisb, line 25: This setting is not defined here: https://edx.readthedocs.io/projects/edx-developer-docs/en/latest/micro_frontends/index.html#required-environment-variables
This means that it's a setting that is specific to the learning MFE, right? If yes, I suggest to add it only to the learning MFE configuration. This is possible by adding `"env": {"production": {"ENABLE_LEARNER_RECORD_MFE": "true"}}` to the "LEARNING_MFE_APP" config entry in plugin.py.

Source:
https://github.com/overhangio/tutor-mfe/blob/5807603a9893d355357d4379ef1b4e634ecc2fd4/tutormfe/templates/mfe/build/mfe/Dockerfile#L59
https://github.com/overhangio/tutor-mfe/blob/5807603a9893d355357d4379ef1b4e634ecc2fd4/tutormfe/plugin.py#L32

ghassanmas: Yup you are totally right... just re pushed

regisb (approved):

regisb:
Thanks!

PR #56, opened 2022-07-11
Remove wrapping quotes from env settings

jcohen28:
Prior versions of docker-compose had a bug that did not properly escape quotation marks in env_files, incorrectly including them as characters in the string value. The issue was fixed in docker-compose version 1.26.0. If tutor is run on a system that uses an earlier version of docker-compose, then tutor-mfe functionality gets broken by invalid config values.

This should hopefully avoid the issues discussed here: https://discuss.openedx.org/t/mfe-failing-login-refresh-with-404-error/7716/6


arbrandes (approved):
Makes sense.  Good catch.

PR #58, opened 2022-08-18
perf: re-order Dockerfile stages to significantly optimize build speed

kdmccormick:
## Description

Previously, build stages were ordered like this in the
rendered Dockerfile (dependencies in parenthases):

* base
* i18n        (base)
* app1-src    (base)
* app1-i18n   (base, app1-src)
* app1-dev    (base, app1-src, app1-i18n)
* app1        (app1-dev)
* app2-src    (base, app2-src)
* app2-i18n   (base, app2-src)
* app2-dev    (base, app2-src, app2-i18n)
* app2        (app2-dev)
* ...
* production  (app1, app2, ..., appN)

MFEs in dev mode only use the output of the *-dev stages.
However, because some of the *-dev stages came after
the intermediate production stages in the Dockerfile (app1, app2, etc),
Docker unnecessarily built the prod stages whenever `tutor dev start`
was run, which often forced users to wait through the prod-only
`npm run build`.

This commit reorders the build stages to look like this:

* base
* i18n        (base)
* app1-src    (base)
* app1-i18n   (base, app1-src)
* app1-dev    (base, app1-src, app1-i18n)
* app2-src    (base, app2-src)
* app2-i18n   (base, app2-src)
* app2-dev    (base, app2-src, app2-i18n)
* ...
* app1        (app1-dev)
* app2        (app2-dev)
* ...
* production  (app1, app2, ..., appN)

Now that the *-dev stages are all before the production stages,
the MFE dev build should be significantly quicker.

## Review

@arbrandes Can you take a look?

### DIFF
--- a/tutormfe/templates/mfe/build/mfe/Dockerfile
+++ b/tutormfe/templates/mfe/build/mfe/Dockerfile
@@ -53,6 +53,11 @@ COPY --from={{ app["name"] }}-i18n /openedx/app/src/i18n/messages /openedx/app/s
 ENV PUBLIC_PATH='/{{ app["name"] }}/'
 EXPOSE {{ app['port'] }}
 CMD ["npm", "run", "start"]
+
+{% endfor %}
+
+{% for app in iter_values_named(suffix="MFE_APP") %}
+
 ######## {{ app["name"] }} (production)
 FROM {{ app["name"] }}-dev AS {{ app["name"] }}
 COPY ./env/production /openedx/env/production
--- a/tutormfe/templates/mfe/build/mfe/Dockerfile
+++ b/tutormfe/templates/mfe/build/mfe/Dockerfile
@@ -24,19 +24,21 @@ RUN echo "copying i18n data" \
   echo "done."
 
 {% for app in iter_values_named(suffix="MFE_APP") %}
+
 ######## {{ app["name"] }} (src)
 FROM base AS {{ app["name"] }}-src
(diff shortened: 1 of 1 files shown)


kdmccormick:
FYI @regisb (when you're back). I realized this issue while working on https://github.com/overhangio/2u-tutor-adoption/issues/87.

ghassanmas:
In the issue I was referring is this part of code: 

https://github.com/overhangio/tutor-mfe/blob/0b695b9e0acdc6a0f3a107f5fcdc07946142a696/tutormfe/templates/mfe/build/mfe/Dockerfile#L48-L49

The part of `rm -rf .npm` I think here probably Regis meant to optimize the image total size one its finish building. I think there is a room for imporvment here, I don't know how exactly, ideally I think it would be: 

When mfe is installing its depds, it can uses .npm cahches of other mfe that are already installed, but however apps can be installed in parrallel, (_if you enable docker buildkit_), so probably if we can have the base base to already cahces common pacakges, then defeinilty would decrease network bandwidth usage (i.e install time),. All of this would more possible or easier to implement when we have or agree on set of depends that are consistant among the mfes, which is I opened a some PRs against nutmeg branch for few mfes.

arbrandes:
@kdmccormick, will review/test ASAP.  Thanks for this!

arbrandes (approved):
The changes make sense.  I also tested the build, and it works fine.  Approved.

regisb:
This is a great contribution, thanks!

PR #59, opened 2022-09-13
fix: escape space in platform name

ghassanmas:
tutor-mfe set SITE_NAME env variable of mfe to equal
  PLATFORM_NAME config value, however in case it has a space, this
  would result SITE_NAME to be empty.
  This change prefix space with \ as, i.e My Site becaomse
  My\ Site.
  Before this might not has been issued because all values were
  wrapped with ' hence #56

  To test this change (you have to be in production mode)
  - Set your PLATFORM_NAME to any thing that incldues space
  - Build the mfe image
  - Check `document.title` in dev console of any mfe
  - It will not incldue PLATFORM_NAME

  After applying this patch, the last step above, should result
  the PLATFORM_NAME i.e. "Learning | PLATFORM_NAME"

### DIFF
--- a/tutormfe/templates/mfe/build/mfe/env/production
+++ b/tutormfe/templates/mfe/build/mfe/env/production
@@ -19,7 +19,7 @@ NODE_ENV=production
 PUBLISHER_BASE_URL=
 REFRESH_ACCESS_TOKEN_ENDPOINT={% if ENABLE_HTTPS %}https{% else %}http{% endif %}://{{ LMS_HOST }}/login_refresh
 SEGMENT_KEY=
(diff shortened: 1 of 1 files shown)


regisb (approved):
Whithout this patch I observe an issue with PLATFORM_NAME="My Open edX". During build I get the following error:

```
Step 97/112 : RUN bash -c "set -a && source /openedx/env/production && source /openedx/env/production.override && npm run build"                                                              
 ---> Running in 6136e7089067                                                                                                                                                                 
/openedx/env/production: line 22: Open: command not found
```

I can confirm that pulling this change fixes the issue.

PR #63, opened 2022-10-11
feat: add learner record MFE to plugin

Tj-Tracy:
Adds the learner record MFE to `tutor-mfe` to use in place of the old UI in credentials, as we are planning on deprecating soon.

image

### DIFF
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -44,6 +44,17 @@ config = {
                 },
             },
         },
+        "RECORDS_MFE_APP": {
+            "name": "records",
+            "repository": "https://github.com/openedx/frontend-app-learner-record",
+            "port": 1990,
+            "env": {
+                "production": {
+                    "SUPPORT_URL_LEARNER_RECORDS":"",
+                    "USE_LR_MFE": "",
+                },
+            },
+        },
     },
 }
 
--- a/README.rst
+++ b/README.rst
@@ -9,6 +9,7 @@ In addition, this plugin comes with a few MFEs which are enabled by default:
 - `Gradebook <https://github.com/edx/frontend-app-gradebook/>`__
 - `Learning <https://github.com/edx/frontend-app-learning/>`__
 - `Profile <https://github.com/edx/frontend-app-profile/>`__
+- `Learner Records <https://github.com/edx/frontend-app-learner-record/>`__
 
 Instructions for using each of these MFEs are given below.
 


regisb:
I don't think that we should merge this in master right now, as it would be a huge change to the Nutmeg release. Maybe we can merge to nightly instead?

Also, could you please add some docs about the learner records MFE to the README? Other MFEs have some screenshots and explanations.

tutormfe/plugin.py
```diff
@@ -44,6 +44,17 @@
...
             },
         },
+        "RECORDS_MFE_APP": {
+            "name": "records",
+            "repository": "https://github.com/openedx/frontend-app-learner-record",
+            "port": 1990,
+            "env": {
+                "production": {
+                    "SUPPORT_URL_LEARNER_RECORDS":"",
+                    "USE_LR_MFE": "",
```
regisb, line 54: Is there a good reason why these settings do not have these values as default upstream?

regisb (changes requested):
Aren't we supposed to make changes to the LMS/CMS learner records configuration in order to point them to the MFE container?

tutormfe/plugin.py
Tj-Tracy: the first settings currently only has an edX-specific default, and the second one is for a feature that currently unsupported in tutor. Would it be a better idea to just leave them out? the MFE still operates if they are not present.

Tj-Tracy:
I'm fine with the nightly release. When I have a moment I will change the target and address feedback, thanks for taking a look at this!

hurtstotouchfire:
> I don't think that we should merge this in master right now, as it would be a huge change to the Nutmeg release. Maybe we can merge to nightly instead?
> 
> Also, could you please add some docs about the learner records MFE to the README? Other MFEs have some screenshots and explanations.

@regisb  Is there not a stable Nutmeg release? From an Open edX release perspective this shouldn't go into Nutmeg as it's being added for Olive, so we should PR against whatever is appropriate for Olive and going forward.

As far as docs, we're working with Jenna to collect some context about the new MFE here: https://github.com/openedx/frontend-app-learner-record/issues/107

regisb:
> @regisb Is there not a stable Nutmeg release? From an Open edX release perspective this shouldn't go into Nutmeg as it's being added for Olive, so we should PR against whatever is appropriate for Olive and going forward.

There is! The Nutmeg releases are built off the master branch. Changes that affect meant for future releases (including Olive) should go to the nightly branch. See this section of the docs for more information: https://docs.tutor.overhang.io/tutorials/nightly.html

tutormfe/plugin.py
regisb: > the first settings currently only has an edX-specific default

Can we change the upstream default such that it works out of the box for everyone instead of just edX? 

> Would it be a better idea to just leave them out?

Not if they are needed, no.

hurtstotouchfire:
We would like to close this PR and follow up instead with instructions in the [LRMFE README](https://github.com/openedx/frontend-app-learner-record/blob/master/README.rst) for how to configure the MFE with Tutor.

tutormfe/plugin.py
regisb: FTR, it was decided in a BTR meeting that instead of adding the learner-records MFE to Tutor (and thus enabling it by default for everyone) there would be instructions on how to enable it, either here in the tutor-mfe README, or in the learner-records MFE README.

PR #64, opened 2022-10-12
feat: enable gzip compression

ghassanmas:
Compressing assests would lead to readuce transfer size.
 As testing with frontend-app-learning/Olive, the network traffic
 before was about ~4MB, after this it became ~1MB.

 This change was suggested by Google Lighthouse[1], there are of
 course more suggestion but this was one the easiest and one of most
 impactful.

 [1]: https://web.dev/uses-text-compression
 
 ## Lighthouse score before: 
 
image

### Suggestion Before 

image

 ## Lighthouse score after: 
 
image

### Suggestions After

image

### DIFF
--- a/tutormfe/patches/caddyfile
+++ b/tutormfe/patches/caddyfile
@@ -3,5 +3,6 @@
     request_body {
         max_size 2MB
     }
+    encode gzip
     import proxy "mfe:8002"
 }


regisb:
This looks like a 100% beneficial change. May I suggest that we add this `encode gzip` directive to the base `proxy` directive in the core Caddy config? https://github.com/overhangio/tutor/blob/7453e70fa08943652f72bb7956872ce44e2b1bdf/tutor/templates/apps/caddy/Caddyfile#L25
When loading the LMS dashboard I get the following improvement : `2.39 MB / 651.91 KB transferred` Without gzip encoding I get `2.39 MB / 2.39 MB transferred`. So it's a ~75% page size decrease!

ghassanmas:
Make sense!, and then both the lms/cms, can benfit from it. 
I will open a PR on tutor, and I will add [link/info](https://caddyserver.com/docs/caddyfile/directives/encode) about the defaults.

ghassanmas:
Yes, it no need for it at the tutor-mfe level... It's already in tutor nutmeg/master but for olive I suppose tutor olive branch would at somepoint to pick commits from master i.e rebased

PR #73, opened 2022-10-27
fix: Remove hardcoded loopback ip in dev docker-compose

### LINKED ISSUE #72: Remove hardcoded loopback ip in dev docker-compose
In the docker compose for development, the loopback IP address is [hardcoded to 127.0.0.1](https://github.com/overhangio/tutor-mfe/blob/103dfb4acd890c64305c3f9fb9575c40f5148509/tutormfe/patches/local-docker-compose-dev-services#L9).
In some dev environments, e.g. in cloud environments like Cloud9, or in Windows WSL, the internal web server is accessed via a different IP address. Removing the 127.0.0.1 in the ports section will fix this problem.

angonz:
In the docker compose for development, the loopback IP address is hardcoded to 127.0.0.1. In some dev environments, e.g. in cloud environments like Cloud9, or in Windows WSL, the internal web server is accessed via a different IP address. Removing the 127.0.0.1 in the ports section will fix this problem.
Fixes #72

### DIFF
--- a/tutormfe/patches/local-docker-compose-dev-services
+++ b/tutormfe/patches/local-docker-compose-dev-services
@@ -6,7 +6,7 @@
         context: ../plugins/mfe/build/mfe/
         target: {{ app["name"] }}-dev
     ports:
-        - "127.0.0.1:{{ app["port"] }}:{{ app["port"] }}"
+        - "{{ app["port"] }}:{{ app["port"] }}"
     stdin_open: true
     tty: true
     volumes:


tutormfe/patches/local-docker-compose-dev-services
```diff
@@ -6,7 +6,7 @@
         context: ../plugins/mfe/build/mfe/
         target: {{ app["name"] }}-dev
     ports:
-        - "127.0.0.1:{{ app["port"] }}:{{ app["port"] }}"
+        - "{{ app["port"] }}:{{ app["port"] }}"
```
regisb, line 9: Unfortunately, this is not really a great idea because it exposes the app port to everyone. (0.0.0.0)

That being said, I now realize that the dev ports are indeed open to everyone for the core Tutor app... https://github.com/overhangio/tutor/blob/master/tutor/templates/dev/docker-compose.yml So I guess it's fine to do the same for the MFE.

regisb (approved):

PR #75, opened 2022-11-05
Fix (build): use NPM_REGISTRY from the config

ARMBouhali:
**TL;DR -** Use the `NPM_REGISTRY` config in Dockerfile, the same way it's done in other Tutor Dockerfiles (tutor-discovery, tutor-ecommerce, tutor:openedx, tutor-contrib-credentials).

We just replace the default registry URL with the config `NPM_REGISTRY` which defaults to the same value.

This is required in order to use a custom registry either for caching or for accessing private packages.

**Testing**
I've been using that fix for a while. I'm currently surviving long repetitive downloads thanks to my local pull-through cache NPM registry powered by [Verdaccio](https://verdaccio.org/docs/what-is-verdaccio/).

FYI @arbrandes @regisb


arbrandes (approved):
Makes total sense to me.  Thanks!

PR #77, opened 2022-11-07
docs: Document that DNS needs to be configured

arbrandes:
As per https://discuss.openedx.org/t/studio-superuser-login-user-account-profiles-and-indigo-theme-issues-with-maple-tutor/7363/5?u=arbrandes

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,10 +22,17 @@ Installation
 Usage
 -----
 
-::
+To enable this plugin, run::
 
     tutor plugins enable mfe
     tutor local quickstart
+
+When running the plugin in production, it is recommended that you set up a catch-all CNAME for subdomains at the DNS provider: see the `Configuring DNS Records <https://docs.tutor.overhang.io/install.html#configuring-dns-records>`__ section in the Tutor documentation for more details.  This way, the plugin will work out of the box with no additional configuration.  Which is to say, if your ``LMS_HOST`` is set to `myopenedx.com` the MFEs this plugin provides will be accessible under `apps.myopenedx.com` by default.
+
+To check what the current value of `MFE_HOST` is actually set to, run the following::
+
+    tutor config printvalue MFE_HOST
+
 
 Account
 ~~~~~~~


README.rst
```diff
@@ -175,6 +175,13 @@ In some cases, for example when using `GitLab's NPM package registry <https://do
             RUN npm config set '//gitlab.example.com/api/v4/projects/<your_project_id>/packages/npm/:_authToken' '<your_token>'
             RUN npm install '@edx/frontend-component-header@npm:@foo/<your_frontend_component_header_name>@latest'
 
+Running in production
+---------------------
+
+When running this plugin in production, in addition to the standard `DNS configuration <https://docs.tutor.overhang.io/install.html#configuring-dns-records>`__ it is necessary for the domain name that corresponds to the ``MFE_HOST`` setting to be properly configured at a registrar.  Once the domain can be resolved at, for example, ``apps.your.domain``, you should run::
```
regisb, line 181: This is not quite right. The DNS config instructions suggest to add a CNAME for all subdomains. This should be enough in the default case but of course it will break in the following cases:

- the users did not read the docs relative to the DNS config (this is very likely as the docs do not make a great job at bubbling up this piece of information). Maybe right below (or in?) the usage instructions?
- the user configures a different `MFE_HOST`.

In any case, the DNS instructions are very important, so they should be higher up in the README. Maybe in the

README.rst
```diff
@@ -175,6 +175,13 @@ In some cases, for example when using `GitLab's NPM package registry <https://do
...
+
+    tutor config save --set MFE_HOST=apps.your.domain
```
regisb, line 183: I would be wary of adding that line. When following along a readme, many users (including me sometimes) just blindly copy-paste any verbatim instructions. I know, I know... but it's a fact of life, I've seen it happen with my own eyes.

Instead, maybe we can indicate that the MFE will be running at the host indicated by:

    tutor config printvalue MFE_HOST

And users would do well to follow the [instructions on setting up the DNS records](https://docs.tutor.overhang.io/install.html#configuring-dns-records).

regisb (changes requested):

README.rst
arbrandes: > This is not quite right. The DNS config instructions suggest to add a CNAME for all subdomains.
> the user configures a different MFE_HOST.

Technically, "properly configured at a registrar" is exactly what this means.  After all, as you mention, the user doesn't *have* to use a CNAME if they don't want to, particularly if they have a different `MFE_HOST` set up.  In any case, for consistency, I'll reword it accordingly.

> Maybe right below (or in?) the usage instructions? (...) higher up in the README.  Maybe in the (?)

Your last comment was cut off at the end, but I'll take the previous placement suggestion. 👍🏼

README.rst
arbrandes: The link to the DNS records is already there, but I agree printing out the value is a better approach.  Will do.

regisb (approved):
Nice.

PR #85, opened 2022-11-15
Delete user account issue on Account MFE

CodeWithEmad:
Hi everyone
I noticed that deleting the user account doesn't work by default.

after a [little search](https://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/configuration/user_retire/service_setup.html#retirement-states), I found that we need to populate retirement states before deleting a user.
I've tested this and it worked after retirement states were populated.
 

this PR will add `populate_retirement_states` command if the Account MFE was selected.

### DIFF
--- a/tutormfe/templates/mfe/tasks/lms/init
+++ b/tutormfe/templates/mfe/tasks/lms/init
@@ -1,5 +1,6 @@
 {% if MFE_ACCOUNT_MFE_APP %}
 (./manage.py lms waffle_flag --list | grep account.redirect_to_microfrontend) || ./manage.py lms waffle_flag --create --everyone account.redirect_to_microfrontend
+./manage.py lms populate_retirement_states
 {% else %}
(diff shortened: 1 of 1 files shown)


regisb (approved):
niiiiiiiiiice

CodeWithEmad:
Thanks @regisb 

btw, if I recall correctly we talked about a change log file for all overhang libraries like tutor-mfe. (right now I think only tutor has such a file)
any updates on that?

arbrandes (approved):
Thanks!

regisb:
> btw, if I recall correctly we talked about a change log file for all overhang libraries like tutor-mfe. (right now I think only tutor has such a file)

Right, thanks for reminding me. The current changelog process is not satisfying, so I propose to use scriv: https://github.com/overhangio/tutor/pull/745 If this goes through, then I'll propose that we apply a similar process to plugins.

PR #69, opened 2022-10-16
MFE dynamic config

ghassanmas:
Related to  kdmccormick/tutor-mfe/pull/1

When building the docker image each MFE was built from the its crossponding PR: 

- openedx/frontend-app-profile/pull/586
- openedx/frontend-app-learning/pull/955
- openedx/frontend-app-account/pull/603. (_Since this already is merged olive.master was used instaed_)
 - openedx/frontend-app-gradebook/pull/253

The built image is pushed at https://hub.docker.com/r/16084536/mfe 

## How to test/USE 
- Install or update `tutor` from overhangio/tutor/pull/734
- Install `tutor-mfe` from this PR/branch 
- Use the built mfe-image:
  - `docker pull 16084536/mfe:olive.runtime.amd.1`
  - `tutor config save --set MFE_DOCKER_IMAGE=docker.io/16084536/mfe:olive.runtime.amd.1`

### Dockerfile (_How did the Dockerfile looked like prior to build?_)
<details>
<summary>Dockerfile source</summary>

```Dockerfile
    
  FROM docker.io/node:16.14-bullseye-slim AS base

RUN apt update \
  && apt install -y git \
    # required for cwebp-bin
    gcc git libgl1 libxi6 make \
    # additionally required for gifsicle, mozjpeg, and optipng (on arm)
    autoconf libtool pkg-config zlib1g-dev \
    # additionally required for node-sass (on arm)
    python g++

RUN mkdir -p /openedx/app /openedx/env
WORKDIR /openedx/app
ENV PATH ./node_modules/.bin:${PATH}

######## i18n strings
FROM base AS i18n
COPY ./i18n /openedx/i18n
RUN chmod a+x /openedx/i18n/*.js
RUN echo "copying i18n data" \
  && mkdir -p /openedx/i18n/account \
  && mkdir -p /openedx/i18n/gradebook \
  && mkdir -p /openedx/i18n/learning \
  && mkdir -p /openedx/i18n/profile \
  echo "done."



######## account (src)
FROM base AS account-src
RUN git clone https://github.com/openedx/frontend-app-account --branch open-release/olive.master --depth 1 .
RUN stat /openedx/app/src/i18n/messages 2> /dev/null || (echo "missing messages folder" && mkdir -p /openedx/app/src/i18n/messages)

######## account (i18n)
FROM base AS account-i18n
COPY --from=account-src /openedx/app/src/i18n/messages /openedx/app/src/i18n/messages
[...]

PR #87, opened 2022-11-18
feat: Add support for the Course Authoring MFE

arbrandes:
The only feature enabled by default is the "Pages & Resources" menu item in Studio.
    
Waffle flags for the new React editors are created but disabled by default.  (The only editor that seems to be fully featured at the moment is the HTML block editor; Video seems to work without a preview, and Problem is just a placeholder.)
    
The new proctoring interface is also disabled, as it requires the `edx-exams` IDA, which Tutor doesn't currently support.


tutormfe/templates/mfe/build/mfe/env/development
```diff
@@ -8,5 +8,8 @@ MARKETING_SITE_BASE_URL=http://{{ LMS_HOST }}:8000
 NODE_ENV=development
 REFRESH_ACCESS_TOKEN_ENDPOINT=http://{{ LMS_HOST }}:8000/login_refresh
 STUDIO_BASE_URL=http://{{ CMS_HOST }}:8001
+{% if MFE_COURSE_AUTHORING_MFE_APP and MFE_LEARNING_MFE_APP %}
```
regisb, line 11: Why does adding this setting depend on `MFE_COURSE_AUTHORING_MFE_APP`? Is it because the `LEARNING_BASE_URL` is only used in the authoring MFE?

regisb (approved):
Looking great! Just one minor question.

Will you want to add this change to the olive branch, too? If yes, I suggest that you merge first in Olive, and then all changes from olive will be merged in nightly after the release.

tutormfe/templates/mfe/build/mfe/env/development
arbrandes: @regisb, that's right.  If you're wondering why I put it here and not in the MFE's own env in the Tutor config, it's because I couldn't readily find a way to access some of the required template variables there.

In any case, once #69 merges this is going into the `[lms|cms]-*-settings` templates.

regisb: :ok_hand:

arbrandes:
@regisb, rebased on the just-merged dynamic config support.  Want to take a second pass?

On nightly vs olive: I'm more used to having new things in the unstable branch (here, nightly) first, and then backporting them to the stable one once they work well.  I find it is easier to track their evolution, this way.

tutormfe/templates/mfe/build/mfe/env/development
```diff
@@ -0,0 +1,15 @@
+DISCOVERY_API_BASE_URL={% if DISCOVERY_HOST is defined %}http://{{ DISCOVERY_HOST }}:8381{% endif %}
```
regisb, line 1: This file should be deleted now that we have dynamic config, right? (same for production file below)

regisb (changes requested):

tutormfe/templates/mfe/build/mfe/env/development
arbrandes: Apologies - of course!  This got past me on the rebase. 😬

arbrandes:
Rebased and retested.  All good!

PR #89, opened 2022-11-23
feat: Add support for the ORA Grading MFE

arbrandes:
The MFE is accessible by instructors in ORA exercises that have explicit staff grading steps.

The corresponding waffle flag is installed and enabled by default.

Depends on:

- [x] https://github.com/openedx/frontend-app-ora-grading/pull/185.


tutormfe/plugin.py
```diff
@@ -44,6 +44,11 @@
                 },
             },
         },
+        "ORA_GRADING_MFE_APP": {
+            "name": "ora-grading",
+            "repository": "https://github.com/edx/frontend-app-ora-grading",
```
arbrandes, line 49: Yes, this one's still in the `edx` org. :/

regisb (approved):
Great!

arbrandes:
For the record, we need [this fix](https://github.com/edx/frontend-app-ora-grading/pull/146) to land before we can merge this.

arbrandes:
This has also been rebased, but at a reduced priority since we haven't gotten the blocking upstream PR merged, yet.

arbrandes:
This is now:

* targetting the Palm branch
* blocked on https://github.com/openedx/frontend-app-ora-grading/pull/183, as ORA grading doesn't currently support runtime config

vunguyen-dmt:
Hi @arbrandes, I've tried ORA grading MFE, there was one issue I found, could you check it in this release.
In Local deployment, when viewing individual submissions, I noticed that file download URLs were relative paths, because of that  it could not be displayed, the url was `https://{MFE_HOST}/openassessment/fileupload/submissions_attachments/...` instead of `https://{LMS_HOST}/openassessment/fileupload/submissions_attachments/...`
I think the problem could be from[ this file](https://github.com/openedx/frontend-app-ora-grading/blob/master/src/components/FilePreview/hooks.js)
```
  const rendererProps = {
    fileName: file.name,
    url: file.downloadUrl,
    onError: stopLoading,
    onSuccess: () => stopLoading(),
  };
```

I'm not sure this is an optimal way to fix the problem, but it could be
```
  const rendererProps = {
    fileName: file.name,
    url: file.downloadUrl && file.downloadUrl.startsWith("/") ? `${baseUrl}${file.downloadUrl}` :  file.downloadUrl,
    onError: stopLoading,
    onSuccess: () => stopLoading(),
  };
```

vunguyen-dmt:
This may not affect files uploaded to S3, only files uploaded to local storage I think.

arbrandes:
@vunguyen-dmt, do you mind opening an issue in https://github.com/openedx/frontend-app-ora-grading describing the problem you observed, including instructions to reproduce?  It's not readily apparent - at least to me - that this is a problem with tutor-mfe.

In any case, I don't think that is a blocker to releasing in Palm, so I'm going to go ahead and merge this.

PR #94, opened 2022-12-08
feat: add support for the authn MFE

arbrandes:
Adds support for the Authn MFE, enabling it by default.
    
This disable enterprise login by default.
    
Co-authored-by: Adolfo R. Brandes <abrandes@tcril.org>


arbrandes:
@regisb, what do you think?  Shall we?  It seems the issue I was seeing before was [a red herring](https://github.com/openedx/frontend-app-authn/issues/692): fixing CSRF also fixed this.

regisb:
If authn works without edx-enterprise, then let's go for it! I think we should include it in Olive. 
We should probably inform the release test group though: all these new MFEs introduce many new pages that were not part of the test plan.

PR #95, opened 2022-12-09
fix: add missing CSRF configuration

arbrandes:
When introducing dynamic config[1], `CSRF_TOKEN_API_PATH` was removed from the MFE build environment and not added to the corresponding Django settings, leading to POST failures across the board.

This addresses the problem by adding `CSRF_TOKEN_API_PATH` back in.

[1] https://github.com/overhangio/tutor-mfe/pull/69


regisb (approved):

arbrandes:
Tested with `tutor local`, fixes all POST problems.

PR #98, opened 2022-12-13
feat: do not build mfe image on `local start`

regisb:
Now that the mfe image is pushed to docker.io, there is no longer any need to re-build every time we run `tutor local start`.

### DIFF
--- a/tutormfe/patches/local-docker-compose-prod-services
+++ b/tutormfe/patches/local-docker-compose-prod-services
@@ -1,8 +1,6 @@
 # MFE
 mfe:
     image: {{ MFE_DOCKER_IMAGE }}
-    build:
-        context: ../plugins/mfe/build/mfe/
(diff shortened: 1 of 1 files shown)


kdmccormick (approved):

arbrandes:
Wait, didn't I [already fix this](https://github.com/overhangio/tutor-mfe/pull/96)?  How did this disappear from Olive?  I'm pretty sure I was rebasing on nightly right up until the last moment.

arbrandes (approved):

kdmccormick:
You merged into Nightly two days ago, which I believe will feed into Palm's master but missed the cut for Olive's master.

arbrandes:
Yeah, I missed the cut.  I mis-remembered having fixed it on Friday, but it's obviously not the case.  Thanks!

PR #100, opened 2022-12-14
fix: pull dev images from docker.io

arbrandes:
Now that we push development images to docker.io, we also need to use the correct URI to fetch them via docker-compose.


regisb (approved):
Niiiiiice

arbrandes:
@regisb, I snuck a README change in here, so it goes out with 15.0.2.  Mind taking another peek?

regisb (approved):

Issue #105, opened 2023-01-04
The password reset link included in emails is invalid


vunguyen-dmt:
The setup process for AUTHN mfe should set
AUTHN_MICROFRONTEND_DOMAIN = {{ MFE_HOST }}/{{ MYMFE_MFE_APP["name"] }}
otherwise reset password link will be: https://none/password_reset_confirm/....

arbrandes:
Huh.  This went totally under the radar.  Indeed, we need [to set a variable for it](https://github.com/openedx/edx-platform/blob/b01cf355a7b9bf395c1e5d5ee6858b179358c4e9/openedx/core/djangoapps/user_authn/views/password_reset.py#L152), but it's not documented anywhere.

PR #108, opened 2023-01-10
fix: The password reset link included in emails is invalid

vunguyen-dmt:
Fix: https://github.com/overhangio/tutor-mfe/issues/105


regisb (commented):
This is looking good, but I did not test the changes myself. Can you confirm you did @vunguyen-dmt? 
Also, can you please rebase and squash your commits on top of the master branch?

PR #113, opened 2023-02-22
Readme: Fix MFE customization example

waza-ari:
The MFE customization example does throw an exception as it is written right now. I've wrapped the two changes in a list and changed `add_item` to `add_items`.


regisb (approved):
Good catch!

arbrandes (approved):
Thank you, @waza-ari!

PR #114, opened 2023-02-22
fix: visual - missing code tags in Readme

waza-ari:
Added some missing code tags to readme which makes it a bit harder to read (markdown vs rst syntax)


regisb (approved):
Thanks!

arbrandes (approved):

PR #120, opened 2023-04-12
feat: upgrade to Palm

regisb:


tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM docker.io/node:16.14-bullseye-slim AS base
+# https://hub.docker.com/_/node/tags
+FROM docker.io/node:16.20-bullseye-slim AS base
```
mtyaka, line 2: MFEs included in palm have all been upgraded to Node 18 now. Can we bump up the node version here?

regisb: Indeed! I managed to build the image locally. Now building on the Palm demo platform.

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -20,60 +24,65 @@ FROM base AS i18n
...
 {% endfor %}
 
 # Production images are last to accelerate dev image building
-{%- for app in iter_values_named(suffix="MFE_APP") %}
-######## {{ app["name"] }} (production)
-FROM {{ app["name"] }}-common AS {{ app["name"] }}-prod
+{%- for app_name, app in iter_mfes() %}
+######## {{ app_name }} (production)
+FROM {{ app_name }}-common AS {{ app_name }}-prod
 ENV NODE_ENV=production
```
ghassanmas, line 86: Cna you we add here support the new feuture 
`ARG USE_PURGECSS=false`
`ENV USE_PURGECESS=$USE_PURGECSS`
So that it can set at build, i.e. `docker build mfe --build-arg USE_PURGECESS=true`
ref: https://github.com/openedx/frontend-build#optimization

regisb: I just tested building the mfe image with `USE_PURGECSS=true` and now I think I know your pain when you build on macOS and Docker eats all your memory.

I'm reluctant to introduce a build-time argument for such a specific use case. Instead, can I suggest we add a `{{ patch("mfe-dockerfile-prod-pre-build") }}` statement?

ghassanmas: Indeed I don't meant to have it by default,  just to have it as env-variable would just it make it easier for people to use, it, _an extra cli arg vs creating a plguin file_. Secondaly regarding resources extensive #125, which is an issue even without using prugeCSS, https://discuss.openedx.org/t/npm-err-code-err-socket-timeout-when-building-mfe-image/10414. Lastly regarding the diff in file size, I would just add that it's not directly about the file size/request size, as it's about perfoamnce; as it's about the time it takes the browser to read compile the file content.

regisb: I really prefer a simple plugin over a new build arg or tutor setting. The new patch would also have the advantage that it could also be used for other purposes.

arbrandes (approved):
Totally approved. 👍🏼

PR #123, opened 2023-05-11
fix: pngquant error during build on linux/arm64

regisb:
Building the mfe Docker image on arm64 triggers the following error in the 'account-prod' stage:

    Cannot find module 'imagemin-pngquant'

This was due to a missing system package.

I find it strange that no one reported this issue before, when there are plenty of users running macOS with arm64. I discovered this issue while setting up a multi-arch build CI environment.

### DIFF
--- a/tutormfe/templates/mfe/build/mfe/Dockerfile
+++ b/tutormfe/templates/mfe/build/mfe/Dockerfile
@@ -4,10 +4,12 @@ RUN apt update \
   && apt install -y git \
     # required for cwebp-bin
     gcc libgl1 libxi6 make \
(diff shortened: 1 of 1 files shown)


regisb:
@arbrandes I'm taking the liberty to merge this and make a new release to fix the arm64 build.

bradenmacdonald:
> I find it strange that no one reported this issue before

I think that the `mfe` images were usually built as AMD64, probably because whatever base image they were using at the time was amd64 only, though it could be some other reason. So at least personally I never tried building them on ARM. But as mentioned on Slack, it does seem to be working now :)

PR #121, opened 2023-04-25
fix: Add account url in dynamic config

ghassanmas:
frontend-component-header since version v.3.5 requires the
   the settings of ACCOUNT_SETTINGS_URL to exists as it uses
   it point to the account MFE.
   - openedx/frontend-component-header/pull/210 PR that add the setting the componenet
   - openedx/wg-build-test-release/issues/268 which this fixes

PR #129, opened 2023-05-26
fix: "account" menu item from MFE menus

### LINKED ISSUE #128: Links to Account in the header are broken
It is reproducible at demo.openedx.overhang.io
image

This is caused by ACCOUNT_SETTINGS_URL config variable ([link](https://github.com/openedx/frontend-component-header/blob/2543926c959c181e73d8fa61fa8baf1f58f8a48f/src/Header.jsx#L63)) not being set by tutor-mfe.

I tried to [fix it](https://github.com/snglth/tutor-mfe/commit/7caccb2d43107d1a0026c9cef77b13cef855d84e) in my fork. Works fine on `dev` environment, but not on `local` one.

regisb:
The "account" menu item could not be clicked in other MFEs, such as the profile.

This is a backport of another fix to Palm:
https://github.com/overhangio/tutor-mfe/pull/121

Close #128.

### DIFF
--- a/changelog.d/20230526_152532_regis.md
+++ b/changelog.d/20230526_152532_regis.md
@@ -1,0 +1,1 @@
+- [Bugfix] Fix un-clickable "account" menu item. (by @ghassanmas and @regisb)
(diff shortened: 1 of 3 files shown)


regisb:
FYI @arbrandes

PR #131, opened 2023-06-14
feat: invalidate build cache on upstream changes

arbrandes:
This will invalidate the image build cache for each MFE whenever:
    
1. A direct URL to the repository's branch refs is provided
    
and
    
2. The head of the branch being built changed between builds.
    
For Github, this is available via [1].  For Gitlab, via [2].
    
[1] https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#get-a-reference
[2] https://docs.gitlab.com/ee/api/branches.html#get-single-repository-branch

### DIFF
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -26,38 +26,47 @@ config = {
 CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
     "authn": {
         "repository": "https://github.com/openedx/frontend-app-authn",
+        "refs": "https://api.github.com/repos/openedx/frontend-app-authn/git/refs/tags",
         "port": 1999,
     },
     "account": {
         "repository": "https://github.com/openedx/frontend-app-account",
+        "refs": "https://api.github.com/repos/openedx/frontend-app-account/git/refs/tags",
         "port": 1997,
     },
     "communications": {
         "repository": "https://github.com/openedx/frontend-app-communications",
+        "refs": "https://api.github.com/repos/openedx/frontend-app-communications/git/refs/tags",
         "port": 1984,
     },
     "course-authoring": {
         "repository": "https://github.com/openedx/frontend-app-course-authoring",
+        "refs": "https://api.github.com/repos/openedx/frontend-app-course-authoring/git/refs/tags",
         "port": 2001,
     },
     "discussions": {
         "repository": "https://github.com/openedx/frontend-app-discussions",
+        "refs": "https://api.github.com/repos/openedx/frontend-app-discussions/git/refs/tags",
         "port": 2002,
     },
     "gradebook": {
         "repository": "https://github.com/openedx/frontend-app-gradebook",
+        "refs": "https://api.github.com/repos/openedx/frontend-app-gradebook/git/refs/tags",
         "port": 1994,
     },
     "learning": {
         "repository": "https://github.com/openedx/frontend-app-learning",
+        "refs": "https://api.github.com/repos/openedx/frontend-app-learning/git/refs/tags",
         "port": 2000,
     },
     "ora-grading": {
         "repository": "https://github.com/openedx/frontend-app-ora-grading",
+        "refs": "https://api.github.com/repos/openedx/frontend-app-ora-grading/git/refs/tags",
         "port": 1993,
     },
     "profile": {
         "repository": "https://github.com/openedx/frontend-app-profile",
+        "refs": "https://api.github.com/repos/openedx/frontend-app-profile/git/refs/tags",
         "port": 1995,
     },
 }
--- a/tutormfe/templates/mfe/build/mfe/Dockerfile
+++ b/tutormfe/templates/mfe/build/mfe/Dockerfile
@@ -33,6 +33,10 @@ RUN echo "copying i18n data" \
 ####################### {{ app_name }} MFE
 ######## {{ app_name }} (git)
 FROM base AS {{ app_name }}-git
+{#- Invalidate the build cache if a change is detected upstream #}
+{%- if app.get("refs") %}
+ADD {{ app["refs"] }}/{{ app.get("version", MFE_COMMON_VERSION) }} /tmp/gitref-{{ app_name }}
+{%- endif %}
 RUN git clone {{ app["repository"] }} --branch {{ app.get("version", MFE_COMMON_VERSION) }} --depth 1 .
 
 ######## {{ app_name }} (src)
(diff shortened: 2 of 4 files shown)


arbrandes:
Just tested this, and BuildKit seems to be smart enough to only rebuild the MFE in question.  (And yes, only when there's an upstream change.)

ghassanmas (commented):
Thanks @arbrandes, I think if I understand correctly this shall ease the pain when/if: 

1.  I am currently using tutor to build an image based on a spesfic branch
    1. Let's assume this occuras at 1:00pm
2. I go on, because the branch is dev mode situation, and I would update it's ref (by either pushing more commits or rewirite head ref-id) on the same branch
   1. Let's assume this occurs at 1:10pm.  
3. Now when I rebuild the image again: 
    1. Assume a new rebuild is at 1:20 Docker would ignore that one of the layer needs to be rebuilt, because by default it compare the branch/name, not their ref commit-id, and thus the image would still depend on 1.i. 
    
Also assmuing my understand is correct for how git/Github interrupt tags vs branch, where tag is expected to be immutable, though it can be mutated, but it's not recommend to do so. And I suppose it's the reason when creating new sub release, we create new tag by incrementing it's last digit. 

In summary this is expected to be useful when working with branchs not tags,  Am I right?  whe the refs suffix is `git/refs/heads` as oppsite to `git/refs/tags`. 

Note: I haven't tested it yet.

arbrandes:
@ghassanmas,

> this is expected to be useful when working with branchs not tags, Am I right?

Correct!  This only makes sense when rebuilding the tutor-mfe image from branches.  I figure it will be most useful on the couple of months between the cutting of master for the next release and the release itself, when we're frequently rebuilding from the release master branch.

But it could also be useful for nightly rebuilds, or for people that have custom branches that get rebuilt often.

Also, while I included it in the patch, there's really no reason to have default values that point to tags in the master branch.  It was more a way to demonstrate syntax; we can remove those if people think this is a good idea.

tutormfe/plugin.py
```diff
@@ -26,38 +26,47 @@
 CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
     "authn": {
         "repository": "https://github.com/openedx/frontend-app-authn",
+        "refs": "https://api.github.com/repos/openedx/frontend-app-authn/git/refs/tags",
```
arbrandes, line 29: These default `refs` lines should probably be removed before merging; they're here just as a basis for discussion.  (Basically, it doesn't make much sense to use this feature with tags, as they're immutable anyway.)

regisb: I think it does make sense to keep these references, because they are also used in the nightly branch, which makes use of a branch, and not a tag.

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -33,6 +33,10 @@ RUN echo "copying i18n data" \
 ####################### {{ app_name }} MFE
 ######## {{ app_name }} (git)
 FROM base AS {{ app_name }}-git
+{#- Invalidate the build cache if a change is detected upstream #}
+{%- if app.get("refs") %}
+ADD {{ app["refs"] }}/{{ app.get("version", MFE_COMMON_VERSION) }} /.gitref-{{ app_name }}
```
regisb, line 38: Minor: can we place this file not at the root, maybe in /tmp?

regisb (approved):
LGTM! Would you like to address the minor comment? Feel free to merge whenever you want.

tutormfe/plugin.py
arbrandes: Ok!  The only thing is that in the nightly branch we need to remember to change `tags` to `branches`:

```suggestion
        "refs": "https://api.github.com/repos/openedx/frontend-app-authn/git/refs/branches",
```

tutormfe/templates/mfe/build/mfe/Dockerfile
arbrandes: Fair enough!

tutormfe/plugin.py
regisb: Hmmm almost. It should be "https://api.github.com/repos/openedx/frontend-app-authn/git/refs/heads/master" in nightly. Thus the `MFE_COMMON_VERSION` should not be appended to that url in nightly (because "nightly != master"). Also, notice it's "heads" and not "branches".

arbrandes:
Re-tested, works fine.  Merging!

regisb:
Hmmmm but this is going to cause the build to fail in nightly, right? When I reviewed the PR I did not realize that the ref url would be different for a branch.

regisb:
This situation is going to be difficult to handle in the Quince upgrade. I'd rather avoid commits that are exclusive to the nightly branch. Instead, I suggest that we switch dynamically from "tags" to "heads" based on the value of `tutormfe.__about__.__version_suffix__`.

Oh, and I expect that some users will also face issues when they manually set `OPENEDX_COMMON_VERSION` to a branch...

arbrandes:
Ah, didn't know about `__version_suffix__`!  Let me give it a shot.

But changing `OPENEDX_COMMON_VERSION` is going to be an issue, indeed.  I wonder (again) if we should just leave default `refs` out of `master`.  That means `nightly` will have an additional commit to optimize rebuilds, but... it does already have a couple of nightly-specific commits anyway. 🤷🏼‍♂️

regisb:
> it does already have a couple of nightly-specific commits anyway.

Does it? We should really try to avoid this situation as much as possible. That being said, I acknowledge that the design of the [nightly/master](https://docs.tutor.overhang.io/tutorials/nightly.html) branching model leaves much to be desired. I'd be happy to hear other ideas.

arbrandes:
Ok: `tutor` has a couple of nightly-specific commits, not `tutor-mfe`.  I'm with you, though.  We should avoid this as much as possible.

As for master vs nightly, the situation is always going to be a bit tricky.  I'm more used to the master-is-unstable paradigm, so that it is the _release_ branch that has branch-specific commits.  But with publishing versioned packages, that wouldn't work very well.

I know of one alterative we can consider.  If nightly were an integration branch, consisting of an octopus merge of several feature branches, then picking which ones to pull into master should be easy later.

For instance, we could have a series of long-lived branches based on master, such as:

* nightly-base: nightly-specific changes that are always needed
* new-feature
* bug-fix-for-master

Then, whenever desired, the nightly branch is recreated via:

```
git checkout nightly
git reset --hard master
git merge bug-fix-for-master new-feature nightly-base
```

The next release branch could be managed as an integration branch, too.  The advantage is that there is always a single source of truth for a particular changeset, and it gets pulled into whichever integration branch requires it.  That way we always know what _shouldn't_ eventually go into master, but at the same time we can build a single artifact for testing.

The disadvantage is that integration branches don't (re)create themselves, and have no memory.  We'd want a script to do the integration process, and it would need to be modified whenever a new branch needs to be included.

PR #134, opened 2023-06-21
feat: allow patching the Dockerfile per MFE

arbrandes:
This should allow managing individual MFE node_modules (such as installing custom dependencies for l10n purposes) without having to fork the corresponding upstream codebases.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -233,6 +233,25 @@ The same applies to installing a custom `brand <https://github.com/openedx/brand
         )
     )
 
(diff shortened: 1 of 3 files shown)


arbrandes:
This should help with custom localization scenarios such as [the one shared by Dimitris](https://discuss.openedx.org/t/problems-applying-customised-mfe-libraries-eg-frontend-platform-to-a-production-open-edx-platform-using-tutor-local/8691/15?u=arbrandes).

@regisb, thoughts?

regisb (changes requested):
Excellent! Can you please add a changelog entry? I think we should also add some information to the README in the "Customising MFEs" section.

arbrandes:
Done!

regisb (approved):
grand!

PR #135, opened 2023-06-27
fix: pick default GH refs API endpoint based on package version

arbrandes:
Picking up where https://github.com/overhangio/tutor-mfe/pull/131 left off.


arbrandes:
@regisb, basically just a conversation starter.  I'm not sure this is the best way to handle this.  (But it could be. 🤷🏼‍♂️)

regisb (approved):
I think this should be good enough for now. We can probably merge this as-is and then work on an improvement later.

arbrandes:
Tested and it's working, so going ahead and merging.

I'm also going to revert the previous nightly fix, and also add the "nightly" version suffix.

regisb:
With hindsight, this change is really super useful, especially in CI.

PR #137, opened 2023-08-02
fix: high memory/cpu usage in dev

regisb:
In development, the tutor-mfe plugin runs a separate container for every MFE. This runs `npm run start` for each of the 9 MFEs, which causes the developer laptops to crash and ice caps to melt.

Here, we adopt a different solution. By default, all MFEs run in a single container, which is the same as the production container (with caddy). Only the Caddyfile is different. This effectively resolves the performance issue.

But to allow developers to hack on invidual MFEs, we also need to run some of them in development mode, with `npm run start`. To do that, we run separately only those MFEs for which the developer has a bind-mounted directory that corresponds to the MFE repository. This allows developers to keep working on MFEs with the help of `tutor mounts ...` command.


brian-smith-tcril (approved):
This looks great! The changes are quite similar to what I have been doing in my work in progress branch. It'd be awesome to get this merged so I can rebase my work on it!

arbrandes (approved):
This is a great improvement.  The code looks good, but I won't have time to actually test it, this week.  Approving, in any case: feel free to merge at your convenience!

johnvente:
Hi @regisb I tried to mount a mfe for example communications but didn't work for development it may be an error mine, I run this:

- `tutor mounts add communications:$(pwd)/frontend-app-communications:/openedx/app`

regisb:
@johnvente please don't use Github issues for feature requests or for support. Instead, you should create a new topic in the official Open edX forum: https://discuss.openedx.org/

PR #145, opened 2023-08-20
docs: square characters replaced with →

CodeWithEmad:
This will fix the same issue we had in [tutor-xqueue](https://github.com/overhangio/tutor-xqueue/pull/19)

Screenshot from 2023-08-20 11-33-04


regisb (approved):
:+1:

PR #153, opened 2023-10-03
fix: remove trailing slash from PUBLIC_PATH env var

ormsbee:
We previously generated the PUBLIC_PATH base for each MFE with a trailing slash, e.g. "/authn/". But the actual links that point to these from the LMS and Studio don't have a trailing slash in them, e.g. "/authn". This commit removes the trailing slash to be consistent.


ormsbee:
(Meant to make it this PR against nightly.)

PR #154, opened 2023-10-03
fix: remove trailing slash from PUBLIC_PATH env var

ormsbee:
We previously generated the PUBLIC_PATH base for each MFE with a trailing slash, e.g. "/authn/". But the actual links that point to these from the LMS and Studio don't have a trailing slash in them, e.g. "/authn". This commit removes the trailing slash to be consistent.

### DIFF
--- a/changelog.d/20231003_133500_dave_remove_trailing_slash.md
+++ b/changelog.d/20231003_133500_dave_remove_trailing_slash.md
@@ -1,0 +1,11 @@
+
+<!--
+Create a changelog entry for every new user-facing change. Please respect the following instructions:
+- Indicate breaking changes by prepending an explosion 💥 character.
+- Prefix your changes with either [Bugfix], [Improvement], [Feature], [Security], [Deprecation].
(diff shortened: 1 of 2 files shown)


ormsbee:
FYI @brian-smith-tcril, @arbrandes

regisb:
Hi Dave! Thanks for the PR. Does this change fix a particular issue, or is it just a cosmetic change?

brian-smith-tcril:
@regisb if this change gets in it should remove the need to add extra trailing slash handling logic in `frontend-platform` now that we've moved to React Router v6 (see the discussion on https://github.com/openedx/frontend-platform/pull/574 for more details)

regisb:
I have no idea what's going on and I should never have asked :sweat_smile: passing it on to you @arbrandes! I'd just like to ask that you create a changelog entry with `scriv create` and to explain in that entry _why_ the change was made (as opposed to _what_ was made). For instance "make it easier and consistent for frontend-patform to parse the `PUBLIC_PATH`" could be a perfectly reasonable changelog description.

arbrandes (approved):
Looks good!  Thanks, Dave!

Issue #151, opened 2023-10-01
Add Makefile to old plugins


CodeWithEmad:
Hi @arbrandes 
It is essential to update old plugins to match the structure of new ones as [cookiecutter-tutor-plugin](https://github.com/overhangio/cookiecutter-tutor-plugin) evolves. I have noticed that the `tutor-mfe` plugin, along with other old plugins, does not have a Makefile. If the absence of a Makefile is not intentional, I would like to add it to this plugin and others.

since this will affect multiple repos, I'll ping @regisb and @kdmccormick too.

regisb:
> It is essential to update old plugins to match the structure of new ones as [cookiecutter-tutor-plugin](https://github.com/overhangio/cookiecutter-tutor-plugin) evolves.

I'm not sure of that. I would definitely not call it "essential". But I agree that there are some features from the cookiecutter that would be super useful in other plugins, including linting and testing.

Making the same changes in many plugins at once is a little painful for non core contributors. I have a convenient local script that allows me to apply the same changes to many repos (it's basically `tmux` with `set-window-option synchronize-panes on`). If you agree I can make the changes myself?

> ps: I see a pre-commit-config.yaml in the [tutor-credentials](https://github.com/overhangio/tutor-credentials). since all plugins should be in sync, should we consider removing the pre-commit config file?

The credentials plugin was moved from its original author @lpm0073. I believe we can remove this file.

CodeWithEmad:
> I would definitely not call it "essential". 

Yes, you're absolutely right. "Useful" is better.

> If you agree I can make the changes myself?

I have made some changes to a few repositories on my local. To help you with adding the Makefile, using tmux with 'synchronize-panes on' would be helpful but it won't be enough to ensure that running 'make test' will succeed. This is mostly due to the 'test-type'. 
If you agree, I would be happy to work on this for all repositories.

regisb:
> If you agree, I would be happy to work on this for all repositories.

Please do! But let's do one complete PR first. In particular, it would be great if you could:

1. Make sure that `make test` is run on GitHub pull requests.
2. Add `make changelog-entry` and `make changelog` targets to the plugin makefiles.

CodeWithEmad:
Okay. do you have any particular repos in mind?

CodeWithEmad:
Quick question:
[...]

PR #155, opened 2023-10-03
Feat: Testing and Linting

CodeWithEmad:
This will add a Makefile, clean the code with isort and black, and fix typing issues.
Closes https://github.com/overhangio/tutor-mfe/issues/151


CodeWithEmad:
@regisb  Should I modify the `gh_refs_path` function so that `"refs": "GITHUB_API" + gh_refs_path()` is replaced by `"refs":  gh_refs_path("openedx/frontend-app-account")`?

regisb:
> Should I modify the gh_refs_path function so that "refs": "GITHUB_API" + gh_refs_path() is replaced by "refs": gh_refs_path("openedx/frontend-app-account")?

Yes, good idea!

>  Also, don't we wanna have make isort and make format run in the test.yml?

I think that's convenient, yes.

regisb (approved):
I think that's a great step forward, and the PR looks good as it is. @arbrandes and @kdmccormick I'm curious to hear what you think?

CodeWithEmad:
> I think that's a great step forward, and the PR looks good as it is.

I'm happy to hear that!

black and isort steps added and gh_refs_path reformatted.

CodeWithEmad:
I have resolved the conflicts that were caused by merging my last pull request.

.github/workflows/test.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          python-version: 3.8
+      - name: Upgrade pip
+        run: python -m pip install --upgrade pip setuptools
+      - name: Install dependencies
+        run: pip install black isort mypy pylint pytest tutor types-docutils types-PyYAML types-setuptools
+      - name: Format with Black
+        run: make format
+      - name: Sort imports with isort
+        run: make isort
```
regisb, line 23: CI should not run `make isort` and `make format`. It should only test that the code is properly formatted with `make test-format` -- and that is covered by `make test`.

tutormfe/plugin.py
```diff
@@ -24,55 +24,65 @@
...
+    Args:
+        owner (str): The name of the Github repository owner, as a string in 'owner' format.
+        name (str): The name of the Github repository, as a string in 'repo' format.
+
+    Returns:
+        str: A string URL to the Github API, pointing to heads if version_suffix is set, tags otherwise.
+
+    """
+    url = f"https://api.github.com/repos/{owner}/{name}/git/refs/"
+    return url + "heads" if __version_suffix__ else url + "tags"
```
regisb, line 39:
```
return f"https://api.github.com/repos/{owner}/{name}/git/refs/{'heads' if __version_suffix__ else 'tags'}"
```

regisb: Also, let's remove the `owner="openedx"` argument from this function.

.github/workflows/test.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Python tests
```
regisb, line 24: This is not just Python tests, but all tests. You should change the name of that step to reflect that.

regisb (changes requested):

.github/workflows/test.yml
CodeWithEmad: Oops. You're absolutely right.

.github/workflows/test.yml
CodeWithEmad: do you have any particular name in mind?

tutormfe/plugin.py
CodeWithEmad: sure.

.github/workflows/test.yml
CodeWithEmad: Changed to Test lint, types, and format

regisb (approved):

CodeWithEmad:
Just a friendly reminder @kdmccormick  @arbrandes

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Upgrade pip
+        run: python -m pip install --upgrade pip setuptools
+      - name: Install dependencies
+        run: pip install black isort mypy pylint pytest tutor types-docutils types-PyYAML types-setuptools
```
kdmccormick, line 19: This command will install the latest version of each of these dependencies. If any of these dependencies releases breaking changes (such as a new default check) then it could cause the PR checks to start failing.

For this reason, Tutor (and most Open edX repos) use a system where they list their versionless dependencies in a ".in" file, and then generate a list of specifically-versioned dependency pins into a ".txt" file. Requirements are then installed from the .txt file (e.g., `pip install -r requirements/development.txt`). In order to keep the versions fresh, the pins in the .txt files are periodically regenerated from the versionless requirements in the .in files using the `make upgrade-requirements` command. [OEP-18](https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0018-bp-python-dependencies.html) dives deep into the details if you're interested.

Can you update this PR to use such a system? You can use Tutor itself as an example. You'll need to:

* Put these testing requirements into [requirements/dev.in](https://github.com/overhangio/tutor/blob/master/requirements/dev.in)
* Add an [`upgrade-requirements` target to the Makefile](https://github.com/overhangio/tutor/blob/f4dc508c1913b818f69259125493109cea6144f8/Makefile#L18) which runs pip-compile on dev.in
* Run `make upgrade-requirements` in order to generate [requirements/dev.txt](https://github.com/overhangio/tutor/blob/master/requirements/dev.txt)
* Install requirements from requirements/dev.txt in this test workflow.

Most repositories have "base.in" and "docs.in" as well, but I think you can ignore those for this PR.

Let me know if you have any questions. Happy to help clarify.

kdmccormick (changes requested):
Thanks for you patience @CodeWithEmad . I think this is a great improvement!

My only concern is that we should pin the development dependencies in order to keep the PR build stable--see my comment below.

.github/workflows/test.yml
CodeWithEmad: > I think this is a great improvement!

Glad to hear that!

I agree with you and I mentioned having such a file [here](https://github.com/overhangio/tutor-mfe/issues/151#issuecomment-1742621681) also, but @regisb suggested keeping all plugins and cookiecutter itself as lean as possible. If everyone agrees, I'm fine with making updates to the PR. I just want to ensure that everyone is on board with the changes.

kdmccormick: Haha, I should have figured you were already thinking about this.

Curious to hear what you think @regisb .

arbrandes: For what it's worth, I think pinning the requirements is as important with plugins as with any other package.

CodeWithEmad: @regisb I assume you missed this thread.

regisb: Sorry about that!

I suggest we pin the requirements in the CI file. Then, once we are happy with how CI works, we will move it out of the plugin repo to a dedicated repository, such that it can be reused across plugins. That way, we will have both the benefit of a reproducible environment, and a lean plugin repository.

CodeWithEmad: OK, so I used the pinned versions of `dev.txt` in tutor (there was no pinned version for `pytest`, so I used the latest).

regisb (approved):
Thanks a lot Emad! This is very useful. I'll let @arbrandes review, squash and merge.

arbrandes (approved):

CodeWithEmad:
@regisb since this was merged, should I create PRs for other old plugins + update the ci.yaml to install tutor[dev] insead of installing different packages??

Issue #161, opened 2023-10-30
GitHub rate limit hit when running MFE image building in CI/CD pipelines


gabor-boros:
It seems that the tutor-mfe plugin [uses the GitHub API 1](https://github.com/overhangio/tutor-mfe/blob/455bd212544e557591cdba0123878daedac6bc3c/tutormfe/templates/mfe/build/mfe/Dockerfile#L38) that is limited to 60 requests per hour for unauthenticated requests (and 5k for authenticated ones).

> Adding credentials to these requests seems to require tutor-mfe PR to add a curl or wget command call instead of the ADD defined in the Dockerfile at the moment. Docker documentation says:

If your URL files are protected using authentication, you need to use RUN wget, RUN curl or use another tool from within the container as the ADD instruction does not support authentication. – [reference](https://docs.docker.com/engine/reference/builder/#add)

Also, this would require adding optional build arguments to the Dockerfile to set the credentials.

-- related forum discussion: https://discuss.openedx.org/t/tutor-mfe-build-github-api-authentication/11542

gabor-boros:
@regisb @arbrandes 

In my opinion a possible solution could be (what I already mentioned in the issue description) to use `RUN curl ...` instead of `ADD` [in the Dockerfile](https://github.com/overhangio/tutor-mfe/blob/master/tutormfe/templates/mfe/build/mfe/Dockerfile#L38). We could have extra arguments for the Dockerfile to define the auth credentials for the call.

arbrandes:
Presuming `curl` or `wget` is already in the image, it _could_ work.  Note the implied doubt: we're trying to second-guess the caching mechanism by forcing the file to be re-fetched via `ADD` at every single run.  I don't know if `RUN` will have the same effect - it pretty much _doesn't_ for the git clone line.

To test the change, you have to re-run the build after pushing a single commit to the target branch.  If the commit makes it into the build, it worked.  If it didn't... it didn't.

gabor-boros:
@arbrandes It probably won't as the `RUN` command will be cached. The only way to work that around (and not disabling the cache) is to pass the commit reference directly to the Dockerfile instead of the branch name (i.e. do the api call right before building the image). That way we would still have the possibility to authenticate the API calls and to have proper layer caching.

FTR: I don't know how many Tutor plugins (or even the main Tutor repo) are using GitHub APIs like this, but the issue should be fixed at all occurrences.

gabor-boros:
[...]

PR #163, opened 2023-11-14
refactor: use MFE refs caching for master branches

gabor-boros:
This PR fixes https://github.com/overhangio/tutor-mfe/issues/161 by using the `--keep-git-dir` flag of the add command. The docker build step is cached if the branch is intact since the last build. However, if the branch is updated, the cache is invalidated and a new build is required -- as expected.

Screenshot 2023-12-04 at 13 38 50

The image above shows that the first build had ditch the previous cache; the second build shows that the image is rebuilt with no changes on the branch, therefore it is cached.

cc: @regisb @arbrandes

### DIFF
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -28,65 +28,41 @@ config = {
     },
 }
 
-
-def get_github_refs_path(name: str) -> str:
-    """
-    Generate a URL to access refs in heads (nightly) or tags (stable) via Github API.
-    Args:
-        name (str): Consisted of the repository owner and the repository name, as a string in 'owner/repo' format.
-
-    Returns:
-        str: A string URL to the Github API, pointing to heads if version_suffix is set, tags otherwise.
-
-    """
-
-    return f"https://api.github.com/repos/{name}/git/refs/{'heads' if __version_suffix__ else 'tags'}"
-
-
 CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
     "authn": {
         "repository": "https://github.com/openedx/frontend-app-authn",
-        "refs": get_github_refs_path("openedx/frontend-app-authn"),
         "port": 1999,
     },
     "account": {
         "repository": "https://github.com/openedx/frontend-app-account",
-        "refs": get_github_refs_path("openedx/frontend-app-account"),
         "port": 1997,
     },
     "communications": {
         "repository": "https://github.com/openedx/frontend-app-communications",
-        "refs": get_github_refs_path("openedx/frontend-app-communications"),
         "port": 1984,
     },
     "course-authoring": {
         "repository": "https://github.com/openedx/frontend-app-course-authoring",
-        "refs": get_github_refs_path("openedx/frontend-app-course-authoring"),
         "port": 2001,
     },
     "discussions": {
         "repository": "https://github.com/openedx/frontend-app-discussions",
-        "refs": get_github_refs_path("openedx/frontend-app-discussions"),
         "port": 2002,
     },
     "gradebook": {
         "repository": "https://github.com/openedx/frontend-app-gradebook",
-        "refs": get_github_refs_path("openedx/frontend-app-gradebook"),
         "port": 1994,
     },
     "learning": {
         "repository": "https://github.com/openedx/frontend-app-learning",
-        "refs": get_github_refs_path("openedx/frontend-app-learning"),
         "port": 2000,
     },
     "ora-grading": {
         "repository": "https://github.com/openedx/frontend-app-ora-grading",
-        "refs": get_github_refs_path("openedx/frontend-app-ora-grading"),
         "port": 1993,
     },
     "profile": {
         "repository": "https://github.com/openedx/frontend-app-profile",
-        "refs": get_github_refs_path("openedx/frontend-app-profile"),
         "port": 1995,
     },
 }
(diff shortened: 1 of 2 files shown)


tutormfe/plugin.py
```diff
@@ -86,6 +74,27 @@ def _add_core_mfe_apps(apps: dict[str, MFE_ATTRS_TYPE]) -> dict[str, MFE_ATTRS_T
...
+    If the repository of the MFE is not of openedx or the branch is not a
+    *.master branch, we are not setting the ref, hence the build won't benefit
+    from build caches.
+    """
+    config = tutor_config.load_minimal(root)
+
+    for mfe, app in iter_mfes():
+        mfe_app_version = app.get("version", config.get("MFE_COMMON_VERSION"))
+        is_openedx_repo = app["repository"].startswith("https://github.com/openedx/")
+        is_master_branch = re.match(r"(^open-release\/\w+\.)?master$", mfe_app_version) is not None
```
regisb, line 91: Caret `^` should be at the beginning of the regexp, right?

    r"^(open-release\/\w+\.)?master$"

tutormfe/plugin.py
```diff
@@ -86,6 +74,27 @@ def _add_core_mfe_apps(apps: dict[str, MFE_ATTRS_TYPE]) -> dict[str, MFE_ATTRS_T
...
+@hooks.Actions.PROJECT_ROOT_READY.add()
+def _modify_mfes(root: str):
+    """
+    Set the GitHub ref for OpenEdX MFEs that are using the *.master branch.
+
+    If the repository of the MFE is not of openedx or the branch is not a
+    *.master branch, we are not setting the ref, hence the build won't benefit
+    from build caches.
+    """
+    config = tutor_config.load_minimal(root)
```
regisb, line 86: This is kind of a hack... I understand why you need to do this, but I really wish we did not have to do that. 

So, let me suggest an alternative solution. In the Dockerfile, write:

```
{% set app_ref = get_app_ref(app, MFE_COMMON_VERSION) %}
{% if app_ref %}
ADD {{ app_ref }} /tmp/gitref-{{ app_name }}
{% endif %}
```

And in plugin.py:

```
def get_app_ref(app, common_version):
   return "" or ...
tutor_hooks.Filters.ENV_TEMPLATE_VARIABLES.add_item(("get_app_ref", get_app_ref))
```

regisb (changes requested):
I wish we would avoid calling the `PROJECT_ROOT_READY` and manually loading the config. Sorry about the back-and-forth, this feature is more complex than I expected...

tutormfe/plugin.py
gabor-boros: Oh, what a typo. Thanks, I intended to have it at the beginning.

arbrandes (commented):
I'm missing something: how does this fix rate limiting if somebody is rebuilding off of master?

gabor-boros:
Reading back the comments, this raises a question for me: How would this solve the issue at all? 

I mean, if the `ADD` command reaches out to GitHub to check if the cache need to be invalidated, we will hit the rate-limits again, but now only if the image is built for the master branches (many times). Right?

I guess we were not thinking about that @regisb. We were discussing to drop the caching like this during our call. Won't it make sense to do so?

regisb:
> I mean, if the ADD command reaches out to GitHub to check if the cache need to be invalidated, we will hit the rate-limits again, but now only if the image is built for the master branches (many times). Right?

Yes, this is correct. My reasoning is that this problem should not happen too often, if at all. And in particular, it should not affect your project. And when it does, you always have the option to create a plugin to remove the `refs`. Am I understanding this correctly?

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -34,10 +34,7 @@ RUN echo "copying i18n data" \
 ######## {{ app_name }} (git)
 FROM base AS {{ app_name }}-git
 {#- Invalidate the build cache if a change is detected upstream #}
-{%- if app.get("refs") %}
-ADD {{ app["refs"] }}/{{ app.get("version", MFE_COMMON_VERSION) }} /tmp/gitref-{{ app_name }}
-{%- endif %}
-RUN git clone {{ app["repository"] }} --branch {{ app.get("version", MFE_COMMON_VERSION) }} --depth 1 .
+ADD --keep-git-dir=true {{ app["repository"] }}#{{ app.get("version", MFE_COMMON_VERSION) }} .
```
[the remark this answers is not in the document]
regisb: Ha, this is the piece of information that we needed. I think it's OK to require a version that is almost one year old. We will have to bump the minimum requirements in Tutor core.

regisb (changes requested):
This is a breaking change, as it removes the "refs"  attribute and requires a higher Docker version. As a consequence, please add a corresponding changelog entry with `scriv create`. Because this is a breaking change, we should only introduce it in the nightly branch, so you should rebase your branch on top of nightly. Finally, please update the README to remove references to "refs".

gabor-boros:
@regisb Sure thing. I'll do it by EOD

arbrandes:
I think this would be good to get in Quince.  Mind making those changes, @gabor-boros?

regisb:
@gabor-boros if you don't have time that's fine too, I can make the change today. We're in a bit of a rush with the upcoming Quince release...

regisb:
Superseded by https://github.com/overhangio/tutor-mfe/commit/c4ed70cf4e25a1c16b6945e2b73af47d82d31000
Thanks a lot for your help @gabor-boros!

PR #164, opened 2023-11-16
[nightly] feat: add support for the Learner Dashboard MFE

arbrandes:
Adds support for the Learner Dashboard MFE.

The MFE itself still has a few pending fixes to make it release-ready (see the list of PRs in [this issue](https://github.com/openedx/platform-roadmap/issues/217)), but we might as well merge it to nightly as we fix it up.


regisb (approved):
I love the smell of fresh microfrontends in the morning!

regisb:
@arbrandes I just discovered that the learner dashboard port clashes with the ecommerce MFE. Both use 1996:
https://github.com/openedx/frontend-app-ecommerce
https://github.com/openedx/frontend-app-learner-dashboard

As a consequence, we can no longer run the ecommerce plugin in nightly. We should change the port of one of these two MFEs in Tutor. Adolfo: which  solution do you suggest? How does the devstack bypass this issue?

arbrandes:
@regisb, the devstack doesn't ship frontend-app-ecommerce out of the box.  Only frontend-app-payment.   And that uses port 1998.

In any case, I don't have a strong opinion on which one should be changed, architecturally speaking.   I will say, though, that whichever one is changed, it's also worth submitting the change upstream (to the MFE repo, devstack, etc).  Let me know if you think it's the Dashboard that should change.

PR #166, opened 2023-11-27
feat: dev dependency

CodeWithEmad:
Having a `dev` dependency in plugins makes it more maintainable for ci jobs. also, developers can install dev dependencies directly inside each package (a better DevEx)


regisb (changes requested):
Let's avoid making many changes that don't really improve the plugin much. Please remove the commits related to code formatting and renaming the init file.

regisb (commented):
thanks Emad!

PR #169, opened 2023-12-05
fix: auto-build 3rd-party MFEs

regisb:
When a 3rd party develops an MFE and bind-mounts it, it is expected that the dev image will be automatically built on `dev launch`. Also, author should be able to build it with `tutor images build mymfe-dev`. This was not the case for plugins that were loaded after the "mfe" plugin. For instance: "partners" is loaded after "mfe".

See discussion:
https://discuss.openedx.org/t/issues-developing-a-non-core-mfe-using-tutor/11855


arbrandes (approved):
+1

PR #172, opened 2023-12-06
fix: specify port for dev server to listen on

michaelwheeler:
Seems like the `PORT` environment variable was removed at some point after [v14.0.2](https://github.com/overhangio/tutor-mfe/blob/v14.0.2/tutormfe/patches/local-docker-compose-dev-services). This left development containers listening at the default port 8080, which is not forwarded to the host.

### DIFF
--- a/tutormfe/patches/local-docker-compose-dev-services
+++ b/tutormfe/patches/local-docker-compose-dev-services
@@ -23,5 +23,7 @@ mfe:
     restart: unless-stopped
     depends_on:
         - lms
+    environment:
+        - "PORT={{ app['port'] }}"
 {%- endif %}
 {%- endfor %}
--- a/changelog.d/20231208_132545_mpwheel_mfe_dev_ports.md
+++ b/changelog.d/20231208_132545_mpwheel_mfe_dev_ports.md
@@ -1,0 +1,1 @@
+- [Bugfix] Specify port for dev server to listen on (by @michaelwheeler)


tutormfe/patches/local-docker-compose-dev-services
```diff
@@ -23,5 +23,7 @@ mfe:
     restart: unless-stopped
     depends_on:
         - lms
+    environment:
+        - "PORT={{ app['port'] }}"
```
arbrandes, line 27: Have you tried specifying a port for the MFE in your plugin.py?  This is supposed to be taken care of by this line in the Dockerfile:

https://github.com/overhangio/tutor-mfe/blob/master/tutormfe/templates/mfe/build/mfe/Dockerfile#L71

It works for the core MFEs, and should also work for plugins.  Like [this one](https://github.com/openedx/openedx-tutor-plugins/blob/main/plugins/tutor-contrib-learner-dashboard-mfe/tutor_learner_dashboard_mfe/plugin.py#L22).

regisb: Michael, since you upgraded your plugin from v14, I wonder if you noticed there is a new way to declare MFE plugins, and the previous one (with `*_MFE_APP`) does not work anymore: https://github.com/overhangio/tutor-mfe#adding-new-mfes

arbrandes: Ah, and a detail that has probably been ignored so far: most MFEs currently sport a `.env.development` file which defines a default `PORT=XXXX`, and as far as I know _that_ gets picked up by the devserver.

regisb: Hmmmmm @arbrandes I think that @michaelwheeler is right. Out `app['port']` should be used to override the PORT in .env.development. If it doesn't, then that's a bug.

arbrandes: Fair enough!

michaelwheeler:
> Can you please add a changelog entry with `scriv create`?

Should be all set.

PR #173, opened 2023-12-06
fix: append trailing slash to PUBLIC_PATH

arbrandes:
This reverts ae60b2db08ef77a5dfb42b0ec9c2671c59af08a7.  We depend heavily on Webpack's `publicPath` for everything to work in MFEs that are hosted in sub-paths.  This variable affects many things, including URLs that are automatically generated for things like static assets or lazily loaded chunks of code.  If the trailing slash is not appended to, say, `/learning`, a React.lazy() will try to fetch this from Caddy, and fail:

`/learning210.3f8d0b2a9dbf3b1f1e69.js`

While this can (and probably should) be dealt with in frontend-platform, we do it here for now.


regisb:
No worries, I'll rebase the Quince branch right on top of nightly after we merge this.

PR #176, opened 2023-12-08
feat: Enable the new discussion sidebar in the Learning MFE

arbrandes:
This is a new potential headliner for Quince.  With this, learners get the ability to start discussions on a sidebar to the right of any unit.  Like so:

image

After this patch, any new course gets this.

Old courses need manual configuration.  For one, the `edX (new)` provider needs to be selected in the Discussions configuration, in the course's "Pages and Resources":

image

There's also some neat additional configuration, either for new or old courses:

image


regisb (approved):
Oh wow that's awesome! Merge ahead!

PR #179, opened 2023-12-20
feat: add patch at the end of the production layer

MoisesGSalas:
# Description

The `mfe-dockerfile-production-final` patch allows you to add instructions to the end of the final layer. This is useful in the case you want to override the default CMD or ENTRYPOINT instructions.

## Additional info

I think the patch will be useful overall, but my use case is as follows:

I've been exploring the idea of pointing [`Public Path`](https://webpack.js.org/guides/public-path/) to a cloudfront endpoint to speed up the MFE asset delivery. The only problem is that I can't reuse the image between environments because the cloudfront endpoint is hardcoded at build time. What I've been playing around is to use a placeholder string as the  `PUBLIC_PATH` and run a startup script in the container that replaces the placeholder with a value set in an environment variable.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -424,6 +424,12 @@ Example: ``mfe-dockerfile-pre-npm-install-learning`` will only apply any instruc
 
 File changed: ``tutormfe/templates/mfe/build/mfe/Dockerfile``
 
+mfe-dockerfile-production-final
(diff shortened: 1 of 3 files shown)


regisb (approved):

regisb:
Thanks for this PR @MoisesGSalas!

PR #182, opened 2024-01-02
feat: add course authoring mfe base url to env vars

rpenido:
## Description

This PR adds the "COURSE_AUTHORING_MFE_BASE_URL" env var pointing to the Course Authoring MFE address.

## Additional Info
We are rendering a course authoring URL in an iFrame inside the library authoring as part of the Tagging Project, so we need to know the Course Authoring MFE address.
- https://github.com/openedx/frontend-app-library-authoring/pull/400

### DIFF
--- a/tutormfe/patches/openedx-lms-development-settings
+++ b/tutormfe/patches/openedx-lms-development-settings
@@ -38,6 +38,7 @@ MFE_CONFIG["ACCOUNT_SETTINGS_URL"] = ACCOUNT_MICROFRONTEND_URL
 {% if get_mfe("course-authoring") %}
 MFE_CONFIG["ENABLE_NEW_EDITOR_PAGES"] = True
 MFE_CONFIG["ENABLE_PROGRESS_GRAPH_SETTINGS"] = True
+MFE_CONFIG["COURSE_AUTHORING_MFE_BASE_URL"] = "http://{{ MFE_HOST }}:{{ get_mfe("course-authoring")["port"] }}"
 {% endif %}
 
 {% if get_mfe("discussions") %}
--- a/changelog.d/20240108_150445_romulo_add_course_authoring_mfe_base_url.md
+++ b/changelog.d/20240108_150445_romulo_add_course_authoring_mfe_base_url.md
@@ -1,0 +1,1 @@
+- [Improvement] Adds the "COURSE_AUTHORING_MFE_BASE_URL" env var pointing to the Course Authoring MFE address. (by @rpenido)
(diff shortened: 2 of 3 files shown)


rpenido:
Done!

tutormfe/patches/openedx-lms-development-settings
```diff
@@ -38,6 +38,7 @@ MFE_CONFIG["ACCOUNT_SETTINGS_URL"] = ACCOUNT_MICROFRONTEND_URL
 {% if get_mfe("course-authoring") %}
 MFE_CONFIG["ENABLE_NEW_EDITOR_PAGES"] = True
 MFE_CONFIG["ENABLE_PROGRESS_GRAPH_SETTINGS"] = True
+MFE_CONFIG["COURSE_AUTHORING_MFE_BASE_URL"] = "http://{{ MFE_HOST }}:{{ get_mfe("course-authoring")["port"] }}"
```
bradenmacdonald, line 41: @rpenido Shouldn't this be `"http://{{ MFE_HOST }}:{{ get_mfe("course-authoring")["port"] }}/course-authoring"` ? That seems more consistent with all of the other MFE `_URL` variables in this file.

Also, there seems to be some inconsistency about whether these have a trailing slash or not, but most of them (except account and learner-dashboard) do not so that's probably the way to go.

rpenido: To give some context, we are using this in the iFrame.
We are using the `COURSE_AUTHORING_MFE_BASE_URL` here: https://github.com/openedx/frontend-app-library-authoring/pull/400/files

We have the url `/course-authoring` in the `COURSE_AUTHORING_MICROFRONTEND_URL` variable already set here:

https://github.com/overhangio/tutor-mfe/blob/19efd6c2d16005b40d3b5d0ce5f3d27d3062505c/tutormfe/patches/openedx-cms-development-settings#L2

But this URL renders an error to me:
`http://apps.local.edly.io:2001/course-authoring/home`
image



The course authoring URL in my tutor instance is the following:
image

Is there some misconfiguration on my side?

bradenmacdonald: It's actually a bug in the course authoring MFE: https://github.com/openedx/frontend-app-course-authoring/pull/784

"Studio Home" _should_ be showing at http://apps.local.overhang.io:2001/course-authoring/home (or edly.io)

So I think `COURSE_AUTHORING_MICROFRONTEND_URL` is all you need, and you can close this PR.

rpenido:
Thank you @bradenmacdonald!
Closing!

Issue #185, opened 2024-02-06
Redirection of MFE_HOST URL to LMS_HOST


hinakhadim:
When a user tries to hit the `http://apps.local.edly.io` URL, it doesn't show anything. Correct User Experience should be redirection of this URL to LMS_HOST i.e. `http://local.edly.io`. As in the given image, after entering the URL, it won't show anything as the URL returns `204`.

<img width="1666" alt="Screenshot 2024-02-06 at 4 52 23 PM" src="https://github.com/overhangio/tutor-mfe/assets/112946189/438819f3-bc3d-444d-9237-7eab65da9f11">

PR #186, opened 2024-02-06
fix: Issue MFE_HOST url redirects to LMS_HOST 

hinakhadim:
This PR resolves the #185 . The MFE_HOST url without any MFE name in url redirects to the LMS_HOST url. Before it, the MFE_HOST responds with 204.

### DIFF
--- a/tutormfe/patches/caddyfile
+++ b/tutormfe/patches/caddyfile
@@ -1,5 +1,5 @@
 {{ MFE_HOST }}{$default_site_port} {
-    respond / 204
+    redir / {% if ENABLE_HTTPS %}https://{% else %}http://{% endif %}{{ LMS_HOST }}
     request_body {
         max_size 2MB
     }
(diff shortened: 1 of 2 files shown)


tutormfe/patches/caddyfile
```diff
@@ -1,5 +1,5 @@
 {{ MFE_HOST }}{$default_site_port} {
-    respond / 204
+    redir / {% if ENABLE_HTTPS %}https://{% else %}http://{% endif %}{{ LMS_HOST }}{$default_site_port}
```
regisb, line 2: I think you need to remove the `{$default_site_port}` part. Otherwise it will redirect to the port that is used internally.

regisb (changes requested):
Please add a changelog entry: https://docs.tutor.overhang.io/tutor.html#contributing

tutormfe/patches/caddyfile
hinakhadim: Yup right, I have tried by removing it. It's working fine.

regisb (approved):
LGTM!

regisb:
@arbrandes what do you think?

PR #193, opened 2024-02-28
fix: dev mode after frontend-build namespace change

arbrandes:
Only affects master development builds, so should go into nightly.  See openedx/axim-engineering#23.


DawoudSheraz (approved):

arbrandes:
Thanks for the review, @DawoudSheraz!  Merging.

PR #195, opened 2024-03-01
feat: add course authoring mfe base url to mfe config

rpenido:
## Description
This PR adds the "COURSE_AUTHORING_MFE_BASE_URL" to the mfe config pointing to the Course Authoring MFE address.

## Additional Info
We are rendering a course authoring URL in an iFrame inside the library authoring as part of the Tagging Project, so we need to know the Course Authoring MFE address.

- https://github.com/openedx/frontend-app-library-authoring/pull/400

### DIFF
--- a/changelog.d/20240301_160746_romulo_add_course_authoring_url_mfe_config.md
+++ b/changelog.d/20240301_160746_romulo_add_course_authoring_url_mfe_config.md
@@ -1,0 +1,1 @@
+- [Improvement] Adds the "COURSE_AUTHORING_MFE_BASE_URL" to MFE_CONFIG pointing to the Course Authoring MFE address. (by @rpenido)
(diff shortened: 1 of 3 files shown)


arbrandes:
Shouldn't this be part of a (or the) Library Authoring plugin?  There's an experimental one [here](https://github.com/openedx/openedx-tutor-plugins/tree/main/plugins/tutor-contrib-library-authoring-mfe).

rpenido:
> Shouldn't this be part of a (or the) Library Authoring plugin? There's an experimental one [here](https://github.com/openedx/openedx-tutor-plugins/tree/main/plugins/tutor-contrib-library-authoring-mfe).

As this config refers to the course-authoring url, I thing it is better to set it alongside with the others course-authoring configs. Also, if another MFE also needs it, it don't need to be set again in another plugin.

What do you think?

bradenmacdonald:
@arbrandes  I initially thought the same thing, because the Library Authoring MFE is currently the only MFE that uses this setting, but I lean toward putting it here. I think it makes more sense architecturally for the Course Authoring MFE to "announce itself" and tell other MFEs where to find it than for the Library Authoring MFE plugin to include code about how to find the Course Authoring MFE and to apply a variable about the Course Authoring MFE to all other MFEs. I'm fine with either approach though!

arbrandes (approved):
I like the MFE announcing itself idea.  It's a better solution than having an if statement for each MFE that would require the URL.  It also standardizes on variable names.

PR #196, opened 2024-03-05
fix: add missing curly braces

MoisesGSalas:
# Description

The `openedx-lms-development-settings` was missing the closing braces and caused an error when rendering.


arbrandes (approved):
Thank you!

PR #202, opened 2024-04-11
fix: convert variables that replace mfe .env into string

felipemontoya:
This PR converts values from Boolean to String in the cases where the value replaces `.env` values used in MFEs.

It is one of the options for solving https://github.com/openedx/frontend-app-course-authoring/issues/835

### DIFF
--- a/tutormfe/patches/openedx-lms-production-settings
+++ b/tutormfe/patches/openedx-lms-production-settings
@@ -28,7 +28,7 @@ MFE_CONFIG = {
 {% if get_mfe("authn") %}
 AUTHN_MICROFRONTEND_URL = "{% if ENABLE_HTTPS %}https://{% else %}http://{% endif %}{{ MFE_HOST }}/authn"
 AUTHN_MICROFRONTEND_DOMAIN  = "{{ MFE_HOST }}/authn"
-MFE_CONFIG["DISABLE_ENTERPRISE_LOGIN"] = True
+MFE_CONFIG["DISABLE_ENTERPRISE_LOGIN"] = "true"
 {% endif %}
 
 {% if get_mfe("account") %}
@@ -37,8 +37,8 @@ MFE_CONFIG["ACCOUNT_SETTINGS_URL"] = ACCOUNT_MICROFRONTEND_URL
 {% endif %}
 
 {% if get_mfe("course-authoring") %}
-MFE_CONFIG["ENABLE_NEW_EDITOR_PAGES"] = True
-MFE_CONFIG["ENABLE_PROGRESS_GRAPH_SETTINGS"] = True
+MFE_CONFIG["ENABLE_NEW_EDITOR_PAGES"] = "true"
+MFE_CONFIG["ENABLE_PROGRESS_GRAPH_SETTINGS"] = "true"
 MFE_CONFIG["COURSE_AUTHORING_MICROFRONTEND_URL"] = "{% if ENABLE_HTTPS %}https://{% else %}http://{% endif %}{{ MFE_HOST }}/course-authoring"
 {% endif %}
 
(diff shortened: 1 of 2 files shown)


tutormfe/patches/openedx-lms-production-settings
```diff
@@ -72,7 +72,7 @@ MFE_CONFIG["ACCOUNT_PROFILE_URL"] = "{% if ENABLE_HTTPS %}https://{% else %}http
 
 {% if get_mfe("communications") %}
 COMMUNICATIONS_MICROFRONTEND_URL = "{% if ENABLE_HTTPS %}https://{% else %}http://{% endif %}{{ MFE_HOST }}/communications"
-MFE_CONFIG["SCHEDULE_EMAIL_SECTION"] = True
+MFE_CONFIG["SCHEDULE_EMAIL_SECTION"] = "true"
```
hinakhadim, line 75: Why are we replacing bool to string here? We are not comparing anywhere it with "true" [here](https://github.com/search?q=repo%3Aopenedx%2Ffrontend-app-communications++SCHEDULE_EMAIL_SECTION&type=code).

tutormfe/patches/openedx-lms-production-settings
```diff
@@ -28,7 +28,7 @@ MFE_CONFIG = {
 {% if get_mfe("authn") %}
 AUTHN_MICROFRONTEND_URL = "{% if ENABLE_HTTPS %}https://{% else %}http://{% endif %}{{ MFE_HOST }}/authn"
 AUTHN_MICROFRONTEND_DOMAIN  = "{{ MFE_HOST }}/authn"
-MFE_CONFIG["DISABLE_ENTERPRISE_LOGIN"] = True
+MFE_CONFIG["DISABLE_ENTERPRISE_LOGIN"] = "true"
```
hinakhadim, line 31: Same here. Can you please share why are we converting to string here as well? I checked this [here](https://github.com/search?q=repo%3Aopenedx%2Ffrontend-app-authn%20DISABLE_ENTERPRISE_LOGIN&type=code). There is no comparison done against a string.

tutormfe/patches/openedx-lms-production-settings
felipemontoya: Thanks a lot for the review. I changed all the bools with strings in this PR because they would all trip with the same issue with `process.env` returning strings and MFE_CONFIG returning bool.
I might have overreached as I did not test all of them.

Given the latest round of comments in the course-authoring PR I think we might close this PR altogether. I'll update this PR after we settle the discussion there.

hinakhadim: I have gone through the course-authoring discussion and just got that the issue was in Quince 3. Is that resolved now so I can close this PR?

arbrandes: The reason for replacing all bool to strings is that this would be the _theoretically_ correct behavior for all MFE_CONFIG variables.  I'll reproduce what I wrote on another PR:

> `getConfig()` still returns values as they were set by the .env files, unless they were overriden by runtime config. Since .env files are still supported, and the dotenv package that reads them in doesn't support booleans, we have no choice but to assume everything that `getConfig()` spits out is a string.

The problem is that some MFEs don't follow this assumption correctly.  For instance, to use your example search, take a look here:

https://github.com/openedx/frontend-app-communications/blob/9279a3e4ce888f6cc198ecb63c0638bf8a795667/src/components/bulk-email-tool/bulk-email-form/BulkEmailForm.jsx#L307

This line is actually wrong.  The correct check would be ```getConfig().SCHEDULE_EMAIL_SECTION === "true"```.  What it's actually doing now is just checking to see if the variable is empty of not.  Setting it to "False" in a .env file would not have the expected results.

All of this said, we'll just have to live with some inconsistencies in tutor-mfe until everything conforms to the same standard upstream.

felipemontoya:
Given that https://github.com/openedx/frontend-app-course-authoring/pull/951 was merged with a different approach and it removed one of the keys I was modifiying in this PR, I think the best course of action for this PR is to close it.
For the consistency of other flags we can open a distinct PR for each case.

PR #206, opened 2024-05-01
fix: serve authoring static files in dev mode

arbrandes:
The new Studio implementation in MFE-land requires rendering XBlocks in iframes.  If we want it to work in dev mode, we need to change Tutor's webpack dev config to match upstream's (as per https://github.com/openedx/frontend-app-course-authoring/pull/964/files#diff-03651daa986b59c8421cba91a0ff6e7ffdfc40623e571f6525f648c2a6e19308).

This is draft mode until that PR is merged.


arbrandes:
This is actually not needed, since Tutor inherits the webpack dev config from upstream for each MFE.

PR #216, opened 2024-06-20
feat: introduces MFE_DOCKER_IMAGE_DEV_PREFIX

rediris:
Introduces a `MFE_DOCKER_IMAGE_DEV_PREFIX` setting to customize dev MFE image docker repos. See discussion below for the benefits.

### DIFF
--- a/changelog.d/20240621_123709_rediris_customize_dev_docker_image_tag.md
+++ b/changelog.d/20240621_123709_rediris_customize_dev_docker_image_tag.md
@@ -1,0 +1,1 @@
+- [Feature] Introduce a `MFE_DOCKER_IMAGE_DEV_PREFIX` setting to customize dev MFE image docker repos. (by @rediris)
--- a/tutormfe/patches/local-docker-compose-dev-services
+++ b/tutormfe/patches/local-docker-compose-dev-services
@@ -9,8 +9,12 @@ mfe:
 {%- for app_name, app in iter_mfes() %}
 {%- set mounts = iter_mounts(MOUNTS, app_name)|list %}
 {%- if mounts %}
+{%- set base_docker_image = DOCKER_IMAGE_OPENEDX %}
+{%- if ":" in base_docker_image %}
+{%- set base_docker_image = DOCKER_IMAGE_OPENEDX.split(':')[0] %}
+{%- endif %}
 {{ app_name }}: # Work on this MFE for development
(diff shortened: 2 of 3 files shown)


rediris:
@regisb thoughts?

regisb:
I'm not sure I understand. What problem exactly is this PR solving? Please describe your use case precisely, and how this change improves your workflow.

rediris:
Yes, this probably deserves a bit more explanation than what I provided above.

Scenario 1: when customizing existing MFEs
Using a custom docker repo for MFE development images is currently impossible because the overhangio images are hard-coded into the plugin. While it is possible to override the `tutor local` mode production MFE image by using `MFE_DOCKER_IMAGE` in `config.yml`, this does not apply to the individual MFE images when running tutor in dev mode. This creates a bit of a development bottleneck, as the dev is forced to manually build/rebuild the development images instead of expediently pulling them from the custom docker repo. Additionally, the current process seems to necessitate using `--no-cache` and/or `--no-registry-cache` flags to prevent automatically pulling the default overhangio images.

Scenario 2: when creating new MFEs
The hard-coded overhangio image references create a similar problem when creating new MFEs. Let's say I have a custom MFE called `frontend-app-customizeme` - the resulting docker image path will look like 
`docker.io/overhangio/openedx-customizeme-dev:17.0.1` (based on what I'm seeing here: https://github.com/overhangio/tutor-mfe/blob/master/tutormfe/plugin.py#L154-L155).

Let me know if any of that makes sense...

regisb:
OK I understand. For once, the "right" fix for this is to introduce a new setting: `MFE_DOCKER_IMAGE_DEV_PREFIX="{{ DOCKER_REGISTRY }}overhangio/openedx-"`. Can you please change your PR to follow this approach? Please remember to include a changelog entry: https://docs.tutor.overhang.io/tutor.html#contributing

rediris:
> OK I understand. For once, the "right" fix for this is to introduce a new setting: `MFE_DOCKER_IMAGE_DEV_PREFIX="{{ DOCKER_REGISTRY }}overhangio/openedx-"`. Can you please change your PR to follow this approach? Please remember to include a changelog entry: https://docs.tutor.overhang.io/tutor.html#contributing

Will do, thank you very much!

rediris:
@regisb all set, let me know what you think!

rediris:
Oops, forgot the changelog...

regisb (approved):
lgtm. @arbrandes what do you think?

hinakhadim:
I have tested it. LGTM!

DawoudSheraz (approved):

PR #219, opened 2024-07-26
feat: faster init by avoiding repeated waffle flag listing

regisb:
Calling `./manage.py lms waffle_flag --list` so many times causes the MFE initialisation step to take a very long time, because `./manage.py lms` is very slow. We avoid this by listing existing waffle flags in a temporary file.

On an already initialized system, `tutor local do init --limit=mfe`
drops from 3min 27s to 25s.


DawoudSheraz (approved):
If you have some before and after stats, please add them to PR. 
Looks good.

rediris (approved):

regisb:
I have added some stats to the git message.

PR #225, opened 2024-09-20
feat: rename "course-authoring" MFE to "authoring"

### LINKED ISSUE #224: Rename `course-authoring` -> `authoring`?
frontend-app-course-authoring was renamed to frontend-app-authoring: https://github.com/openedx/frontend-app-authoring/pull/1289

At a minimum, we should update the URLs here.

Should we rename the app as well? It would be a breaking change for plugin authors.

If so, can we add a `/course-authoring/*` -> `/authoring/*` redirect to caddy to avoid breaking links?

regisb:
Existing URLs are (301, permanently) redirected to the new ones. We are doing this because the course-authoring MFE has been migrated to a new repository.

Close #224.


DawoudSheraz:
Shouldn't this be merged to nightly? The name changes in the current release can break behaviors if folks are using custom MFE plugins.

regisb:
You are most definitely right Dawoud.

arbrandes (approved):
Works for me!

PR #230, opened 2024-10-29
v18.0.2

regisb:


regisb:
I'm in a bit of a rush, so I'll merge this now if you don't mind. The goal is to get the latest redwood.3 benefits to the image that is automatically pulled from docker hub. I'll do the same for the Indigo plugin.

PR #232, opened 2024-10-31
fix: add script to install node_modules in mounted mfes

### LINKED ISSUE #218: Mounted MFEs with node_modules missing fails to start the server
This issue is a similar issue to https://github.com/overhangio/tutor/issues/997. The steps to reproduce are:

- Installed latest tutor master or nightly
- Clone any MFE repo (authn, learner-dashboard, etc.) and check out its appropriate branch (open-release/redwood.master if latest tutor master, master if tutor nightly)
- Add the cloned MFE as a bind mount
- Re-build mfe-dev images
- Run platform. Verify the mounted-MFE container runs into the error `fedx-script not found`
- Run npm install within cloned MFE repo on the host. Once installed, MFE dev container should be up and running.

Like the mentioned issue in tutor core, the bind-mounted MFE during image build does not get node_modules copied back to host. When running tutor, the host MFE directory without node_modules overrides the installed node_modules within image, thus resulting in fedx-script not found. 

- Add a similar fix as core issue i.e. if node_modules are not found in MFE directory, run npm clean-install to install the required dependencies
  - If there is another fix, feel free to suggest it.

hinakhadim:
#### WIP

These code changes introduce improvements to the Tutor MFE (Micro-Frontend) plugin by adding automation and efficiency for managing and running tasks in environments where MFEs are mounted. When MFEs are mounted, it will run `npm clean-install` as part of the run `tutor dev launch`. The "mfe" init task `npm-install.sh` is taking care of installing the node_modules, as it is done in the [LMS](https://github.com/overhangio/tutor/blob/master/tutor/templates/jobs/init/mounted-directories.sh)

Test Cases:
- [x] It will run the `npm-install` script when MFEs are **implicitly** and **expicitly** mounted using the Tutor format: `tutor mounts add $(pwd)/frontend-app-*` & `tutor mounts add [mfe-name]:$(pwd)/frontend-app-*:/openedx/app`.
- [x] The job should run for all mounted MFEs
- [ ] The job will run only in `tutor dev`, not `tutor local`, as MFEs mount only in `tutor dev` environment.
- [x] If no `node_modules` are present, it should install `node_modules`.
- [x] If the `node_modules` directory exists but is empty or outdated, it should install `node_modules`.
- [x] If the `package-lock.json` file changes slightly, it should trigger reinstallation of `node_modules`.
- [x] If `node_modules` are installed and then reinstallation is attempted without changes to `package-lock.json`, it should **not** reinstall `node_modules`.

Closes #218

### DIFF
--- a/tutormfe/patches/dev-docker-compose-jobs-services
+++ b/tutormfe/patches/dev-docker-compose-jobs-services
@@ -1,0 +1,11 @@
+{%- for app_name, app in iter_mfes() %}
+{%- set mounts = iter_mounts(MOUNTS, app_name)|list %}
+{%- if mounts %}
+{{ app_name }}-job: 
+    image: "{{ MFE_DOCKER_IMAGE_DEV_PREFIX }}-{{ app_name }}-dev:{{ MFE_VERSION }}"
+    volumes:
+        {%- for mount in mounts %}
+        - {{ mount }}
+        {%- endfor %}
+{%- endif %}
+{%- endfor %}
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import functools
 import os
+import re
 import typing as t
 from glob import glob
 
@@ -304,3 +305,27 @@ def _check_mfe_host(config: Config) -> None:
             f'Warning: MFE_HOST="{mfe_host}" is not a subdomain of LMS_HOST="{lms_host}". '
             "This configuration is not typically recommended and may lead to unexpected behavior."
         )
+    
+
+@tutor_hooks.Actions.CONFIG_LOADED.add()
+def _run_jobs_in_mounted_mfes(config: Config) -> None:
+    mounts = get_typed(config, "MOUNTS", list, [])
+
+    pattern = re.compile(r'frontend-app-(\w+)')
+    mounted_mfes = [match.group(1) for mount in mounts if (match := pattern.search(mount))]
+
+    mfe_task_file = os.path.join(
+        str(
+            importlib_resources.files("tutormfe")
+            / "templates"
+            / "mfe"
+            / "tasks"
+            / "mfe"
+            / "npm-install.sh"
+        )
+    )
+
+    for mounted_mfe in mounted_mfes:
+        with tutor_hooks.Contexts.app("mfe").enter():
+            with open(mfe_task_file, encoding='utf-8') as fi:
+                tutor_hooks.Filters.CLI_DO_INIT_TASKS.add_item((mounted_mfe , fi.read()))
--- a/tutormfe/templates/mfe/tasks/mfe/npm-install.sh
+++ b/tutormfe/templates/mfe/tasks/mfe/npm-install.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Paths for the fingerprint and package-lock.json
+FINGERPRINT_FILE="node_modules/.fingerprint"
+PACKAGE_LOCK="package-lock.json"
+
+# Function to generate the SHA1 hash of the package-lock.json file
+generate_fingerprint() {
+  sha1sum "$PACKAGE_LOCK" | awk '{print $1}'
+}
+
+# Check if node_modules exists and fingerprint is valid
+if [ ! -d "node_modules" ]; then
+  echo "⚠️  node_modules not found! Installing dependencies..."
+  npm clean-install
+
+elif [ ! -f "$FINGERPRINT_FILE" ]; then
+  echo "⚠️  Fingerprint file not found! Re-installing dependencies..."
+  npm clean-install
+
+else
+  CURRENT_FINGERPRINT=$(generate_fingerprint)
+  STORED_FINGERPRINT=$(cat "$FINGERPRINT_FILE")
+
+  if [ "$CURRENT_FINGERPRINT" != "$STORED_FINGERPRINT" ]; then
+    echo "⚠️  package-lock.json changed! Re-installing dependencies..."
+    npm clean-install
+  else
+    echo "✅ Dependencies are up to date."
+    exit 0
+  fi
+fi
+
+# Store the new fingerprint after installation
+generate_fingerprint > "$FINGERPRINT_FILE"
+echo "✅ Dependencies installed and fingerprint updated."


tutormfe/plugin.py
```diff
@@ -304,3 +305,27 @@ def _check_mfe_host(config: Config) -> None:
             f'Warning: MFE_HOST="{mfe_host}" is not a subdomain of LMS_HOST="{lms_host}". '
             "This configuration is not typically recommended and may lead to unexpected behavior."
         )
+    
+
+@tutor_hooks.Actions.CONFIG_LOADED.add()
+def _run_jobs_in_mounted_mfes(config: Config) -> None:
+    mounts = get_typed(config, "MOUNTS", list, [])
+
+    pattern = re.compile(r'frontend-app-(\w+)')
```
Danyal-Faheem, line 314: We can use the REPO_PREFIX variable defined above here instead of a hardcoded string.

tutormfe/plugin.py
```diff
@@ -304,3 +305,27 @@ def _check_mfe_host(config: Config) -> None:
             f'Warning: MFE_HOST="{mfe_host}" is not a subdomain of LMS_HOST="{lms_host}". '
             "This configuration is not typically recommended and may lead to unexpected behavior."
         )
+    
+
+@tutor_hooks.Actions.CONFIG_LOADED.add()
+def _run_jobs_in_mounted_mfes(config: Config) -> None:
+    mounts = get_typed(config, "MOUNTS", list, [])
+
```
Danyal-Faheem, line 313: We can return here if no mounts exist in the config.

tutormfe/patches/dev-docker-compose-jobs-services
```diff
@@ -0,0 +1,11 @@
+{%- for app_name, app in iter_mfes() %}
```
DawoudSheraz, line 1: why is this patch needed?

tutormfe/plugin.py
DawoudSheraz: +1

tutormfe/plugin.py
```diff
@@ -304,3 +305,27 @@ def _check_mfe_host(config: Config) -> None:
...
+    pattern = re.compile(r'frontend-app-(\w+)')
+    mounted_mfes = [match.group(1) for mount in mounts if (match := pattern.search(mount))]
+
+    mfe_task_file = os.path.join(
```
DawoudSheraz, line 317: mfe_npm_install could be a better name as this action is related only to npm install.

tutormfe/templates/mfe/tasks/mfe/npm-install.sh
```diff
@@ -0,0 +1,36 @@
...
+# Paths for the fingerprint and package-lock.json
+FINGERPRINT_FILE="node_modules/.fingerprint"
+PACKAGE_LOCK="package-lock.json"
+
+# Function to generate the SHA1 hash of the package-lock.json file
+generate_fingerprint() {
+  sha1sum "$PACKAGE_LOCK" | awk '{print $1}'
+}
+
+# Check if node_modules exists and fingerprint is valid
```
DawoudSheraz, line 12: this comment can be broken down and mentioned against the appropriate if-elif-else block.

tutormfe/plugin.py
```diff
@@ -304,3 +305,27 @@ def _check_mfe_host(config: Config) -> None:
             f'Warning: MFE_HOST="{mfe_host}" is not a subdomain of LMS_HOST="{lms_host}". '
             "This configuration is not typically recommended and may lead to unexpected behavior."
         )
+    
+
+@tutor_hooks.Actions.CONFIG_LOADED.add()
+def _run_jobs_in_mounted_mfes(config: Config) -> None:
```
regisb, line 311: I'm not a big fan of adding callbacks to the CONFIG_LOADED action. It means we have to manually parse `MOUNTS` and this is bound to fail in many edge cases. Also, we have to start using Context objects, which is VERY dirty.

Instead, I suggest the following:

```
@tutor_hooks.Filters.CLI_DO_INIT_TASKS.add()
def _mfe_npm_install(tasks):
    for mfe in get_mfes():
        if hooks.Filters.COMPOSE_MOUNTS.apply([], f"frontend-app-{mfe}"):
            tasks.append((mfe, "... contents of npm-install.sh..."))
```

What do you think?

tutormfe/templates/mfe/tasks/mfe/npm-install.sh
```diff
@@ -0,0 +1,46 @@
...
+    # Everything is up to date
+    echo "✅ Dependencies are up to date."
+    exit 0
+  fi
+fi
+
+
+# Store the new fingerprint after installation
+generate_fingerprint > "$FINGERPRINT_FILE"
+echo "✅ Dependencies installed and fingerprint updated."
```
Danyal-Faheem, line 46: nit: add extra line at end of file

tutormfe/patches/dev-docker-compose-jobs-services
```diff
@@ -0,0 +1,11 @@
...
+{%- set mounts = iter_mounts(MOUNTS, app_name)|list %}
+{%- if mounts %}
+{{ app_name }}-job: 
+    image: "{{ MFE_DOCKER_IMAGE_DEV_PREFIX }}-{{ app_name }}-dev:{{ MFE_VERSION }}"
+    volumes:
+        {%- for mount in mounts %}
+        - {{ mount }}
+        {%- endfor %}
+{%- endif %}
+{%- endfor %}
```
Danyal-Faheem, line 11: nit: add extra line at end of file

tutormfe/patches/dev-docker-compose-jobs-services
```diff
@@ -0,0 +1,11 @@
+{%- for app_name, app in iter_mfes() %}
+{%- set mounts = iter_mounts(MOUNTS, app_name)|list %}
```
Danyal-Faheem, line 2: We moved from iter_mounts to a cleaner MFEMountData object in #241. We should use the similar format here for consistency.

tutormfe/patches/local-docker-compose-jobs-services
```diff
@@ -0,0 +1,7 @@
+{%- for app_name, app in iter_mfes() %}
```
Danyal-Faheem, line 1: Why is this patch necessary? Don't we just need to run the init job in dev?

Danyal-Faheem (changes requested):

tutormfe/plugin.py
```diff
@@ -304,3 +305,30 @@ def _check_mfe_host(config: Config) -> None:
...
+    mounts = get_typed(config, "MOUNTS", list, [])
+
+    if not mounts:
+        return None
+
+    pattern = re.compile(rf'{re.escape(REPO_PREFIX)}(\w+)')
+    mounted_mfes = [match.group(1) for mount in mounts if (match := pattern.search(mount))]
```
Danyal-Faheem, line 318: Looking back, we should be able to use MFEMountData here as well instead of using a regex. What do you think?

tutormfe/patches/local-docker-compose-jobs-services
```diff
@@ -0,0 +1,14 @@
...
+{%- for app_name, app, mounts in mfe_data.mounted %}
+{{ app_name }}-job:
+    image: "{{ MFE_DOCKER_IMAGE_DEV_PREFIX }}-{{ app_name }}-dev:{{ MFE_VERSION }}"
+    volumes:
+        {%- for mount in mounts %}
+        - {{ mount }}
+        {%- endfor %}
+    environment:
+        - "NODE_ENV=production"
+{%- endfor %}
```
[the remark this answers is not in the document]
Danyal-Faheem: Right, we tried to find a way to do that, however we are using the CONFIG_LOADED action to parse mounts and add our jobs. This action does not provide us any information if the job is being run with the local or dev context and therefore, we can't add it conditionally.

The only hook that provides this information is the `APP_PUBLIC_HOSTS` filter and we cannot use it as it doesn't provide us information regarding mounts. This is why we had no choice but to run the job in both local and dev containers and then exit if a production ENV is discovered.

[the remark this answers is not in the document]
DawoudSheraz: @arbrandes Ping 🙂

Issue #199, opened 2024-03-13
Support Javascript configuration


arbrandes:
### Description

tutor-mfe doesn't currently support [Javascript configuration](https://github.com/openedx/frontend-platform/blob/master/docs/decisions/0007-javascript-file-configuration.rst), except potentially by the old, abandoned expedient of baking it into the image.  The latter is obviously not something we want to go back to, so a way must be devised to take advantage of JS configuration dynamically.

#### Why should we support it?

Because plugins will [require it](https://github.com/openedx/frontend-plugin-framework/?tab=readme-ov-file#host-mfe-js-based-configuration).

### Solution Space

A list of potential solutions, for discussion.

#### 1. Extend mfe_config

Extend [the mfe_config API](https://github.com/openedx/edx-platform/blob/master/lms/djangoapps/mfe_config_api/docs/decisions/0001-mfe-config-api.rst#decision) so that JS configuration can also be served.  Extend [runtimeConfig()](https://github.com/openedx/frontend-platform/blob/master/src/initialize.js#L180) so that it makes use of it.
 
* Pros: uses the existing flow (no need to rethink the architecture)
* Cons: storing Javascript in a string in `site_configuration` doesn't sound like the best idea ever.  Editing it will be horrible, for one.

#### 2. Refactor mfe_config

This is mostly the same as 1., except instead of piggy-backing on `site_configuration`, we write a dedicated admin page for it.

* Pros: less horrible editing experience
* Cons: more work

#### 3. Move mfe_config to tutor-mfe

_Note: this last option is the most similar to how Piral does - or can do - configuration.  I suggest we explore it seriously, as it might make our jobs easier later._

Instead of relying on an edx-platform API, we build a mechanism in tutor-mfe to serve the correct JS file dynamically.  We'd likely be able to leverage the fact that `.env.config.js` is just Javascript to make a request to a predictable location that Caddy can then service.  This file would be user-overridable, so if the operator wants to do something fancy like fetching a different configuration depending on the domain, they totally could.

Pros: 
 * The most flexible option.  An operator could go as far as writing their own configuration backend, and then writing a JS config to use it.
 * Potentially the best performance: no need to deal with Django.
 * Very easy to implement for tutor-mfe developers.
 * For basic use, actually easier to modify MFE configuration than writing plugins to change MFE_CONFIG.  You just edit the JS file.

Cons:
 * For the more complicated things like multi-tenancy, puts the brunt of the work on operators.
 * Unknown territory.

[...]

PR #233, opened 2024-11-07
docs: Add info on how to use frontend plugin slots

arbrandes:
It's now possible to configure frontend plugin slots across MFEs (notably, header and footer).  This documents how this can be done as simply as possible.
    
See #199

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -308,6 +308,122 @@ In case you need to run additional instructions just before the build step you c
 
 You can find more patches in the `patch catalog <#template-patch-catalog>`_ below.
 
+Using Frontend Plugin Slots
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``mfe-dockerfile-pre-npm-build`` patch is also suitable for taking advantage of the ``env.config.jsx`` mechanism, which lets you define and use frontend plugins.  For instance, you can use it to modify the footer across all MFEs *without* having to maintain a fork of ``frontend-component-footer``.
+
+Start by following the Tutor plugin tutorial at the `Adding new templates <https://docs.tutor.edly.io/tutorials/plugin.html#adding-new-templates>`_ section.  Instead of adding a whole new ``Dockerfile``, though, you'll just modify the exiting one.  Your ``myplugin.py`` should look something like this:
+
+.. code-block:: python
+
+    import os
+    from tutor import hooks
+
+    # Add the template root folder
+    template_folder = os.path.join(os.path.dirname(__file__), "templates")
+    hooks.Filters.ENV_TEMPLATE_ROOTS.add_item(template_folder)
+
+    # env.config.jsx files should live in `mfe/build/mfe`, which is also where the
+    # tutor-mfe Dockerfile goes.  This is so it's more straightforward to copy it
+    # in, as seen below.
+    hooks.Filters.ENV_TEMPLATE_TARGETS.add_item(("mfe/build/mfe", "plugins"))
+
+    # This patch gets inserted into the tutor-mfe Dockerfile.  It copies our
+    # env.config.jsx into the root of each and every MFE before the npm build step.
+    hooks.Filters.ENV_PATCHES.add_item(("mfe-dockerfile-pre-npm-build",
+        "COPY env.config.jsx /openedx/app"
+    ))
+
+As for ``env.config.jsx``, as noted above, it should be put into a ``templates/mfe/build/mfe`` subdirectory below ``myplugin.py``.  This is so it's easy to ``COPY`` it into the Docker image.
+
+In order for ``env.config.jsx`` to work with all current MFEs (including ones that don't yet have any plugin slots), it needs to be defined in a particular way.  This is what it should look like:
+
+.. code-block:: javascript
+
+    /* We can't just assume FPF exists, as it's not declared as a dependency in all
+     * MFEs.  Therefore, we have to use dynamic imports to check if it's available.
+     *
+     * We also have to use a `try` block (as opposed to `.then()`), otherwise
+     * Webpack won't treat the import as optional. Hence the async function.
+     */
+    async function getConfig () {
+      let config = {};
+
+      try {
+        const { DIRECT_PLUGIN, PLUGIN_OPERATIONS } = await import('@openedx/frontend-plugin-framework');
+
+        config = {
+          pluginSlots: {
+            footer_slot: {
+              plugins: [
+                {
+                  /* Hide the default footer. */
+                  op: PLUGIN_OPERATIONS.Hide,
+                  widgetId: 'default_contents',
+                },
+                {
+                  /* Insert a custom footer. */
+                  op: PLUGIN_OPERATIONS.Insert,
+                  widget: {
+                    id: 'custom_footer',
+                    type: DIRECT_PLUGIN,
+                    RenderWidget: () => (
+                      <h1 style={{ "{{textAlign: 'center'}}" }}>This is the footer.</h1>
+                    ),
+                  },
+                },
+              ]
+            }
+          }
+        };
+
+        /* NPM allows us to introspect the package name at build time, which in turn
+         * lets us tailor the configuration for individual MFEs.
+         */
+        if (process.env.npm_package_name == '@edx/frontend-app-profile') {
+          config.pluginSlots.footer_slot.plugins[1].widget.RenderWidget = () => (
+            <h1 style={{ "{{textAlign: 'center'}}" }}>This is the profile MFE's footer.</h1>
+          );
+        }
+      } catch { };
+
+      return config;
+    }
+
+    export default getConfig;
+
+As you can see, by using the ``process.env.npm_package_name`` variable you can use a single ``env.config.jsx`` to selectively apply configuration to different MFEs.
+
+If, on the other hand, you prefer to maintain entirely separate ``env.config.jsx`` files, one for each mfe, that can also be done.  For that you can rely on the ``mfe-dockerfile-pre-npm-build-*`` patches.  For instance:
+
+.. code-block:: python
+
+    hooks.Filters.ENV_PATCHES.add_items([
+        (
+            "mfe-dockerfile-pre-npm-build",
+            "COPY env.config.jsx /openedx/app"
+        ),
+        (
+            "mfe-dockerfile-pre-npm-build-account",
+            "COPY env-account.config.jsx /openedx/app/env.config.jsx"
+        ),
+        (
+            "mfe-dockerfile-pre-npm-build-profile",
+            "COPY env-profile.config.jsx /openedx/app/env.config.jsx"
+        )
+    ])
+
+In this case, there's a global ``env.config.jsx``, but it would get overwritten by the specific ones for the Account and Profile MFEs.  Note that while they originally have different names, they should all still be copied to ``env.config.jsx``.
+
+You can find what slots a library or MFE supports by inspecting its ``src/plugin-slots`` directory on Github.  For example:
+
+- `Header slots <https://github.com/openedx/frontend-component-header/tree/master/src/plugin-slots>`_
+- `Learning MFE slots <https://github.com/openedx/frontend-app-learning/tree/master/src/plugin-slots>`_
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -308,6 +308,122 @@ In case you need to run additional instructions just before the build step you c
...
+In order for ``env.config.jsx`` to work with all current MFEs (including ones that don't yet have any plugin slots), it needs to be defined in a particular way.  This is what it should look like:
+
+.. code-block:: javascript
+
+    /* We can't just assume FPF exists, as it's not declared as a dependency in all
+     * MFEs.  Therefore, we have to use dynamic imports to check if it's available.
+     *
+     * We also have to use a `try` block (as opposed to `.then()`), otherwise
+     * Webpack won't treat the import as optional. Hence the async function.
+     */
```
arbrandes, line 349: Note to reviewer: this particular workaround will no longer be necessary once all supported MFEs have at least one defined slot, forcing them to add `frontend-plugin-framework` to their list of dependencies.

README.rst
```diff
@@ -308,6 +308,122 @@ In case you need to run additional instructions just before the build step you c
...
+                  /* Insert a custom footer. */
+                  op: PLUGIN_OPERATIONS.Insert,
+                  widget: {
+                    id: 'custom_footer',
+                    type: DIRECT_PLUGIN,
+                    RenderWidget: () => (
+                      <h1 style={{ "{{textAlign: 'center'}}" }}>This is the footer.</h1>
+                    ),
+                  },
+                },
```
arbrandes, line 375: Note to reviewer: I considered adding `env.config.jsx` templates directly to tutor-mfe and then exposing a new set of patches that would reduce some of the boilerplate, but the trade-off in flexibility for convenience seems to me to be unfavorable.  To take full advantage of frontend plugins, particularly more complicated ones, users will need access to the whole `env.config.jsx` file anyway.

README.rst
```diff
@@ -308,6 +308,122 @@ In case you need to run additional instructions just before the build step you c
...
+    import os
+    from tutor import hooks
+
+    # Add the template root folder
+    template_folder = os.path.join(os.path.dirname(__file__), "templates")
+    hooks.Filters.ENV_TEMPLATE_ROOTS.add_item(template_folder)
+
+    # env.config.jsx files should live in `mfe/build/mfe`, which is also where the
+    # tutor-mfe Dockerfile goes.  This is so it's more straightforward to copy it
+    # in, as seen below.
```
arbrandes, line 329: I realize it's iffy to recommend a plugin put a file directly into one of tutor-mfe's own ones, but for some reason I couldn't get the t utor-mfe Dockerfile to COPY from a relative parent.

README.rst
hinakhadim: Could you please clarify your approach here? I may not have fully grasped the patches thing and the trade-offs you're mentioning regarding flexibility and boilerplate reduction.

README.rst
hinakhadim: In [Indigo](https://github.com/overhangio/tutor-indigo/blob/master/tutorindigo/plugin.py#L50), I implemented a similar setup, but I have some concerns regarding the inclusion of `env.config.jsx` in mfe/build/mfe. Specifically, if we go this route, each user will need to take similar steps as I did in Indigo. This means they’ll either have to fork tutor-mfe and modify `env.config.jsx` or copy `env.config.jsx` from their custom plugin into `mfe/build/mfe`.

Can we think of any workaround to update or change `env.config.jsx` more seamlessly without requiring each user to fork plugin or override this file? 

OR I am unsure that this `env.config.jsx` file will raise any issue for[ indigo plugin](https://github.com/overhangio/tutor-indigo/blob/master/tutorindigo/plugin.py#L134).

README.rst
arbrandes: @hinakhadim, sure, I replied more fully to Regis here:  https://github.com/overhangio/tutor-mfe/pull/233#discussion_r1834559527

README.rst
arbrandes: Ah, I didn't look at how indigo does it.  Let me check.

But yeah, the fact multiple plugins may want to modify env.config.jsx is not something I took into account, here.

README.rst
```diff
@@ -308,6 +308,122 @@ In case you need to run additional instructions just before the build step you c
...
+    hooks.Filters.ENV_TEMPLATE_TARGETS.add_item(("mfe/build/mfe", "plugins"))
```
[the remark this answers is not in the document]
arbrandes: I just realized I didn't address one major point that @hinakhadim hammered home afterwards.  That multiple plugins may want to modify env.config.jsx.  I have to think some more about this.

README.rst
brian-smith-tcril: And even more so with [`frontend-base`](https://github.com/openedx/frontend-base/) - which will include/replace FPF

README.rst
```diff
@@ -308,6 +308,122 @@ In case you need to run additional instructions just before the build step you c
...
+                  widgetId: 'default_contents',
+                },
+                {
+                  /* Insert a custom footer. */
+                  op: PLUGIN_OPERATIONS.Insert,
+                  widget: {
+                    id: 'custom_footer',
+                    type: DIRECT_PLUGIN,
+                    RenderWidget: () => (
+                      <h1 style={{ "{{textAlign: 'center'}}" }}>This is the footer.</h1>
```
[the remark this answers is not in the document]
brian-smith-tcril: I figure just having
```html
<h1>This is the footer.</h1>
```
should be fine here. The main point is being able to paste in some code and see something change. The text doesn't need to be centered to accomplish that.

README.rst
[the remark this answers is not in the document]
arbrandes: I just went over [Indigo's `env.config.jsx`](https://github.com/overhangio/tutor-indigo/blob/nightly/tutorindigo/templates/indigo/env.config.jsx), and I guess that's as good an example as any of my first point above. :)

The question I'm ruminating is: how do we let plugins do all that in a single file and still allow multiple plugins to coexist there?

Maybe the answer is that we should _not_ allow plugin authors to run wild with `env.config.jsx`.  Maybe all they can do is npm-install a plugin package via the `post-npm-install` patch,  import the components via a new `mfe-env-config-imports` patch, then declare them via `mfe-env-config-plugin-slots`.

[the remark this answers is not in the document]
brian-smith-tcril: Multiple plugins modifying `env.config.jsx` is definitely a tricky problem, and the fact that [`tutor-indigo` already leverages it](https://github.com/overhangio/tutor-indigo/blob/nightly/tutorindigo/templates/indigo/env.config.jsx) makes it quite the pressing issue.

The first thought that came to my mind was providing a way for plugins to "add their changes to the pile" so to speak. I tested the following `env.config.jsx` file

```jsx
import { DIRECT_PLUGIN, PLUGIN_OPERATIONS } from '@openedx/frontend-plugin-framework';

const config = {
  pluginSlots: {
    desktop_header_slot: {
      keepDefault: false,
      plugins: [
        {
          op: PLUGIN_OPERATIONS.Insert,
          widget: {
            id: 'custom_header_component_one',
            type: DIRECT_PLUGIN,
            RenderWidget: () => (
              <h1>1</h1>
            ),
          },
        },
      ]
    },
    desktop_header_slot: {
        keepDefault: false,
        plugins: [
        {
          op: PLUGIN_OPERATIONS.Insert,
          widget: {
            id: 'custom_header_component_two',
            type: DIRECT_PLUGIN,
            RenderWidget: () => (
              <h1>2</h1>
            ),
          },
        },
      ]
    },
  },
}

export default config;
```

and found that FPF handles it somewhat gracefully - the second component (the one that renders `2`) is shown and the first component (the one that renders `1`) is not.

In a world where each tutor plugin *only* touches one FPF slot this could be helpful - we could provide a way to "add changes to the pile" and have operators make sure the plugins are doing so in their preferred order. The obvious problem with that, however, is tutor plugins that utilize multiple FPF slots. If plugins A and B both replace the header and footer, and a site operator wants the header from A and footer from B this falls apart.

I'll keep thinking about this. It is definitely not a simple problem.

arbrandes:
@regisb, @brian-smith-tcril, @hinakhadim, based on your feedback, I've tried another approach:

https://github.com/overhangio/tutor-mfe/pull/234

Please review at your convenience. ;)

README.rst
hinakhadim: @brian-smith-tcril , Thank you for the clarification on the role of frontend-base. To confirm my understanding for future development, frontend-base will function as a shell that eventually replaces the `frontend-plugin-framework` (FPF), as well as standard elements like the header and footer.

My anticipation is that, even with `frontend-base` in use, we will still have the flexibility to update the header and footer packages individually. This would mean that packages such as the header, footer, and brand could still be customized and maintained as custom packages while being included in `frontend-base`.

README.rst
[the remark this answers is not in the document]
hinakhadim: Thanks for the productive discussion and @brian-smith-tcril  for testing.

I’m currently facing a challenge with the scenario you mentioned: "If plugins A and B both replace the header and footer, and a site operator wants the header from A and the footer from B, this falls apart."

There was a similar instance where a user needed to use both tutor-indigo and a second plugin for additional customizations. From what I understand, the priority of each plugin’s components is determined by the order of activation. If tutor-indigo is enabled first, followed by the second plugin, the components from the second plugin will have priority.
```Dockerfile
...
RUN npm install @edx/frontend-component-header@npm:indigo-header
RUN npm install @edx/frontend-component-footer@npm:indigo-footer

RUN npm install @edx/frontend-component-header@npm: custom-header
RUN npm install @edx/frontend-component-footer@npm: custom-footer
...
```

 Conversely, if the second plugin is enabled first, tutor-indigo components will take precedence (means we use indigo header/footer in react js project).
```Dockerfile
...
RUN npm install @edx/frontend-component-header@npm: custom-header
RUN npm install @edx/frontend-component-footer@npm: custom-footer

RUN npm install @edx/frontend-component-header@npm: indigo-header
RUN npm install @edx/frontend-component-footer@npm: indigo-footer
...
```

In cases where the operator needs the header from tutor-indigo (assuming that user cannot alter plugin) and the footer from the custom plugin (Plugin B), they would need to adjust Plugin B to exclude its header component, keeping only the footer. This way, Plugin A (Indigo) provides the header, and Plugin B supplies the footer without conflicts.

```Dockerfile
RUN npm install @edx/frontend-component-header@npm: indigo-header
RUN npm install @edx/frontend-component-footer@npm: indigo-footer

RUN npm install @edx/frontend-component-footer@npm: custom-footer
```

Any further insights would be greatly appreciated!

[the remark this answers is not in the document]
brian-smith-tcril: > I’m currently facing a challenge with the scenario you mentioned: "If plugins A and B both replace the header and footer, and a site operator wants the header from A and the footer from B, this falls apart."

It is my understanding that this scenario is handled fully in https://github.com/overhangio/tutor-mfe/pull/234

README.rst
```diff
@@ -308,6 +308,122 @@ In case you need to run additional instructions just before the build step you c
 
 You can find more patches in the `patch catalog <#template-patch-catalog>`_ below.
 
+Using Frontend Plugin Slots
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``mfe-dockerfile-pre-npm-build`` patch is also suitable for taking advantage of the ``env.config.jsx`` mechanism, which lets you define and use frontend plugins.  For instance, you can use it to modify the footer across all MFEs *without* having to maintain a fork of ``frontend-component-footer``.
+
+Start by following the Tutor plugin tutorial at the `Adding new templates <https://docs.tutor.edly.io/tutorials/plugin.html#adding-new-templates>`_ section.  Instead of adding a whole new ``Dockerfile``, though, you'll just modify the exiting one.  Your ``myplugin.py`` should look something like this:
```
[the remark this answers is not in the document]
regisb: (reading this 2 months late, as I somehow missed the notification...)

I had no idea about this debate! I've been correcting people's double spaces in Tutor docs and elsewhere for years now, in a silent manner that must have seemed very passive-aggressive... Thanks for educating me!

PR #234, opened 2024-11-11
feat: support frontend plugins via env.config.jsx

arbrandes:
This provides a mechanism to configure frontend plugins using patches to a base `env.config.jsx`, in such a way that multiple plugins can take advantage of the file (including for purposes beyond frontend plugins) without clobbering each other.

This is an alternate approach to #233.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -308,6 +308,117 @@ In case you need to run additional instructions just before the build step you c
 
 You can find more patches in the `patch catalog <#template-patch-catalog>`_ below.
 
+Using Frontend Plugin Slots
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It's possible to take advantage of this plugin's patches to configure frontend plugin slots. Let's say you want to replace the entire footer with a simple message. Where before you might have had to fork ``frontend-component-footer``, the following is all that's currently needed:
+
+.. code-block:: python
+
+    from tutor import hooks
+
+    hooks.Filters.ENV_PATCHES.add_item(
+        (
+            "mfe-env-config-plugin-footer_slot",
+            """
+              {
+                /* Hide the default footer. */
+                op: PLUGIN_OPERATIONS.Hide,
+                widgetId: 'default_contents',
+              },
+              {
+                /* Insert a custom footer. */
+                op: PLUGIN_OPERATIONS.Insert,
+                widget: {
+                  id: 'custom_footer',
+                  type: DIRECT_PLUGIN,
+                  RenderWidget: () => (
+                    <h1>This is the footer.</h1>
+                  ),
+                },
+              },
+    """
+        )
+    )
+
+Let's take a closer look at what's happening here, starting with the patch name: the prefix is ``mfe-env-config-plugin-`` followed by the name of the slot you're trying to configure. In this case, ``footer_slot``.  (There's a full list of `possible slot names <#mfe-env-config-plugin-slot-catalog>`_ below.) Next, we're hiding the default contents of the footer with a ``PLUGIN_OPERATIONS.Hide``. (Refer to the `frontend-plugin-framework README <https://github.com/openedx/frontend-plugin-framework/#>`_ for a full description of the possible plugin types and operations.) The ``default_contents`` widget ID always refers to what's in an unconfigured slot. Finally, we use ``PLUGIN_OPERATIONS.Insert`` to add our custom JSX component, composed of a simple ``<h1>`` message.  We give it a widgetID of ``custom_footer``.
+
+That's it!  If you rebuild the ``mfe`` image after enabling the plugin, "This is the footer." should appear at the bottom of every MFE.
+
+It's also possible to target a specific MFE's footer. For that, you'd use a patch such as ``mfe-env-config-plugin-profile-footer_slot``, ``profile`` being the name of the MFE and ``footer_slot`` the name of the slot:
+
+.. code-block:: python
+
+    hooks.Filters.ENV_PATCHES.add_item(
+        (
+            "mfe-env-config-plugin-profile-footer_slot",
+            """
+              {
+                /* Hide the global footer we defined earlier. */
+                op: PLUGIN_OPERATIONS.Hide,
+                widgetId: 'custom_footer',
+              },
+              {
+                /* Insert a custom footer specific to the Profile MFE. */
+                op: PLUGIN_OPERATIONS.Insert,
+                widget: {
+                  id: 'custom_profile_footer',
+                  type: DIRECT_PLUGIN,
+                  RenderWidget: () => (
+                    <h1>This is the Profile MFE's footer.</h1>
+                  ),
+                },
+              },
+    """
+        )
+    )
+
+Note that we haven't removed the first ``mfe-env-config-plugin-footer_slot`` patch, so here we hide our ``custom_footer`` instead of ``default_contents``.
+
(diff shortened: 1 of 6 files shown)


arbrandes:
@hinakhadim, I'm particularly interested in your take on whether the provided patches are sufficient for Indigo.

brian-smith-tcril (commented):
Overall this is wonderful! Extremely well documented. Thank you so much!

I left one comment with a small suggestion, and I think one other thing isn't documented here.

We discussed having an "escape hatch" for site operators that want to have total control over the `env.config.jsx` files. I think it'd be nice to add documentation for the "escape hatch" strategy including:
* Where to put `env.config.jsx` files
* How to get the `tutor-mfe` image build to pick up those `env.config.jsx` files
* Downsides to using the "escape hatch" method over the recommended patch method
  * Example: needing to manually add code to `env.config.jsx` files that plugins would add automatically.

hinakhadim:
This is an excellent approach! It enables plugin developers to make adjustments to `env.config.jsx` indirectly, preserving the original file. In essence, it should function well even when multiple plugins are involved and will work well for indigo.  The documentation is very well done, outstanding work!

- If possible, could we include an example of this approach in the documentation for better understanding?
- One consideration: as the list of plugin_slots grows, it might lead to a large number of patches. I’m not entirely sure about the implications of this. In other words, we have to manage the list of plugin_slots. If a plugin_slot name changes, then we have to update it here too. I am assuming it is based on [the point 3 in this conversation.
](https://github.com/overhangio/tutor-mfe/pull/233#discussion_r1834559527)
I’ll test this out and will be happy to share feedback based on testing with tutor-indigo.

README.rst
```diff
@@ -308,6 +308,117 @@ In case you need to run additional instructions just before the build step you c
...
+                  ),
+                },
+              },
+    """
+        )
+    )
+
+Let's take a closer look at what's happening here, starting with the patch name: the prefix is ``mfe-env-config-plugin-`` followed by the name of the slot you're trying to configure. In this case, ``footer_slot``.  (There's a full list of `possible slot names <#mfe-env-config-plugin-slot-catalog>`_ below.) Next, we're hiding the default contents of the footer with a ``PLUGIN_OPERATIONS.Hide``. (Refer to the `frontend-plugin-framework README <https://github.com/openedx/frontend-plugin-framework/#>`_ for a full description of the possible plugin types and operations.) The ``default_contents`` widget ID always refers to what's in an unconfigured slot. Finally, we use ``PLUGIN_OPERATIONS.Insert`` to add our custom JSX component, composed of a simple ``<h1>`` message.  We give it a widgetID of ``custom_footer``.
+
+That's it!  If you rebuild the ``mfe`` image after enabling the plugin, "This is the footer." should appear at the bottom of every MFE.
```
sarina, line 346: Could this be linked to instructions to rebuild the mfe image?

README.rst
```diff
@@ -308,6 +308,117 @@ In case you need to run additional instructions just before the build step you c
...
+        )
+    )
+
+Note that we haven't removed the first ``mfe-env-config-plugin-footer_slot`` patch, so here we hide our ``custom_footer`` instead of ``default_contents``.
+
+If you were to rebuild the MFE image now, the Profile MFE's footer would say "This is the Profile MFE's footer", whereas all the others would still contain the global "This is the footer." message.
+
+.. _mfe-env-config-plugin-slot-catalog:
+
+Here's a list of the currently available frontend plugin slots, separated by category. The documentation of each slot can be found by following the corresponding link.
```
sarina, line 382: What's the strategy to keep this document up to date? Is there a template for new plugin documentation within the MFE that might include a note to update the tutor readme?

sarina (commented):
Reviewed just the README, looks good.

README.rst
arbrandes: I figure the strategy can be similar to the one for new feature and waffle flags in edx-platform.  Which is to say, one of two things happens:

* As soon as a new flag/slot is merged in the upstream repo, somebody also opens a PR to Tutor to support it.  This could indeed benefit from some sort of automated reminder, but given that new MFE features are more and more commonly required to work with PR sandboxes as part of the review process, and that PR sandboxes are built with Tutor, adding these to Tutor nightly will simply be a built-in step in getting the slot merged.

* When Tutor and its plugins are updated when a new new release is coming out, part of the work is combing through upstream repos for added and changed features.  This should become less and less necessary as Tutor nightly gets "automatically" updated as described above, but if by any chance a slot or two were missed, this is where they would be added.

sarina: Sounds good, ty

arbrandes: Oh, and one neat thing about this PR is that it will make it possible for PR sandboxes to configure slots for testing (once the Grove plugin is updated to match).

tutormfe/plugin.py
```diff
@@ -90,12 +127,21 @@ def get_mfes() -> dict[str, MFE_ATTRS_TYPE]:
...
+    return MFE_SLOTS.apply([])
+
+
 @tutor_hooks.Actions.PLUGIN_LOADED.add()
 def _clear_get_mfes_cache(_name: str) -> None:
     """
     Don't forget to clear cache, or we'll have some strange surprises...
     """
     get_mfes.cache_clear()
+    get_slots.cache_clear()
```
regisb, line 144: Let's not do that. This will break in some edge cases. Instead, both `get_slots` and `get_mfes` should use [tutor.hooks.lru_cache](https://docs.tutor.edly.io/reference/api/hooks/index.html#tutor.hooks.lru_cache).

README.rst
regisb: This approach has one major drawback: Implementing a new slot will require making an upstream PR to tutor-mfe: this will cause a lot of back-and-forth, especially as we intend to have many more plugin slots in Open edX.

Another minor drawback is the requirement to create a Tutor plugin for users who implement their own plugin slots.

As far as I understand, there is little benefit to having an explicit list of slots here. All it gives us is some syntax sugar. We can achieve a similar result with the following MFE-specific filter:

```
tutormfe.hooks.Filters.PLUGIN_SLOTS.add_item((
            "profile", "footer_slot",
            """
              {
                /* Hide the global footer we defined earlier. */
                op: PLUGIN_OPERATIONS.Hide,
                widgetId: 'custom_footer',
              }"""
))
tutormfe.hooks.Filters.PLUGIN_SLOTS.add_item((
            "profile", "footer_slot",
              """{
                /* Insert a custom footer specific to the Profile MFE. */
                op: PLUGIN_OPERATIONS.Insert,
                widget: {
                  id: 'custom_profile_footer',
                  type: DIRECT_PLUGIN,
                  RenderWidget: () => (
                    <h1>This is the Profile MFE's footer.</h1>
                  )
                }
              }
))"""
```

regisb (commented):
I like this approach better but I have an issue with the explicit list of plugin slots. Let me know what you think of my suggestion.

README.rst
arbrandes: Yes, but listing all the slots gives us one majoir _advantage_: it will be very evident what slots are actually available for a particular version of Tutor.  Sure, we can still document them here anyway, but if we make it so a slot only works if it's explicitly supported, then it's also obligatory to document a slot when it's added here.

In any case, I'll give the PLUGIN_SLOTS filter a shot.  I'm not sure yet if it's going to work with the single `env.config.jsx` approach I tried here.

arbrandes: Alright, here's the third take: #240

arbrandes:
Closed in favor of #240

PR #238, opened 2024-11-18
fix: the navigation sidebar should be enabled by default

arbrandes:
This is actually an Open edX product requirement for Sumac.


DawoudSheraz:
I have changed the base branch to sumac as this should be merged to sumac and not nightly. You might want to rebase the branch locally before merge.

DawoudSheraz (approved):

arbrandes:
That is a good question.  My gut feeling is that we should let existing users make the change at their own pace as much as possible, but they should also be made aware of the feature.   In other words, we should make sure to note this new behavior in the release notes.

arbrandes:
@regisb, Chelsea reports that Product would actually prefer it if we forced the feature for existing users.  Any objections?  Do we have an example of such a waffle flag flip in a previous upgrade?

regisb:
No problem at all! This is how forum does it: https://github.com/overhangio/tutor/pull/1139/files#diff-574714bb9c86df7c8389e5c0a473db7b023d12f8534b75d7b5d0cc1ef8527967

arbrandes:
Alright, here it is: https://github.com/overhangio/tutor/pull/1176

PR #240, opened 2024-11-20
feat: support frontend plugins via env.config.jsx

arbrandes:
This provides a mechanism to configure frontend plugins using a new PLUGIN_SLOTS filter and a series of patches that modify a base `env.config.jsx`, in such a way that multiple plugins can take advantage of the file (including for purposes beyond frontend plugins) without clobbering each other.

(This is the third take.  Prior: #233, #234)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -308,6 +308,131 @@ In case you need to run additional instructions just before the build step you c
 
 You can find more patches in the `patch catalog <#template-patch-catalog>`_ below.
 
+Using Frontend Plugin Slots
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It's possible to take advantage of this plugin's hooks to configure frontend plugin slots. Let's say you want to replace the entire footer with a simple message. Where before you might have had to fork ``frontend-component-footer``, the following is all that's currently needed:
+
+.. code-block:: python
+
+    from tutormfe.hooks import PLUGIN_SLOTS
+
+    PLUGIN_SLOTS.add_items([
+        # Hide the default footer
+        (
+            "all",
+            "footer_slot",
+            """
+            {
+              op: PLUGIN_OPERATIONS.Hide,
+              widgetId: 'default_contents',
+            }"""
+        ),
+        # Insert a custom footer
+        (
+            "all",
+            "footer_slot",
+            """
+            {
+              op: PLUGIN_OPERATIONS.Insert,
+              widget: {
+                id: 'custom_footer',
+                type: DIRECT_PLUGIN,
+                RenderWidget: () => (
+                  <h1>This is the footer.</h1>
+                ),
+              },
+            }"""
+        )
+    ])
+
+Let's take a closer look at what's happening here.  To begin with, we're using tutormfe's own ``PLUGIN_SLOTS`` filter.  It's a regular Tutor filter, but you won't find it in the main ``tutor`` package:
+
+.. code-block:: python
+
+    from tutormfe.hooks import PLUGIN_SLOTS
+
+Next up, we're adding actual slot configuration, starting by hiding the default footer.  The first parameter in a filter item specifies which MFE to apply the slot configuration to; for example: ``"learner-dashboard"``, or ``"learning"``. We're using ``"all"`` here, which is a special case: it means the slot configuration should be applied to all MFEs that actually have that slot.  (If a particular MFE doesn't have the slot, it will just ignore its configuration.)
+
+The second parameter, ``"footer_slot"``, is the name of the slot as defined in the code of the MFE itself.
+
+.. code-block:: python
+
+    PLUGIN_SLOTS.add_items([
+        # Hide the default footer
+        (
+            "all",
+            "footer_slot",
+            """
+            {
+              op: PLUGIN_OPERATIONS.Hide,
+              widgetId: 'default_contents',
+            }"""
+        ),
+
+The last parameter to ``add_item()`` is a big string with the actual slot configuration, which will be interpreted as JSX. What we're doing there is hiding the default contents of the footer with a ``PLUGIN_OPERATIONS.Hide``. (You can refer to the `frontend-plugin-framework README <https://github.com/openedx/frontend-plugin-framework/#>`_ for a full description of the possible plugin types and operations.) And the ``default_contents`` widget ID we're targetting always refers to what's in an unconfigured slot by default.
+
+In the second filter item, we once again target the ``"footer_slot"`` on ``"all"`` MFEs. This time, we use ``PLUGIN_OPERATIONS.Insert`` to add our custom JSX component, comprised of a simple ``<h1>`` message we're defining in an anonymous function. We give it a widgetID of ``custom_footer``:
+
+.. code-block:: python
+
+    # Insert a custom footer
+    (
+        "all",
+        "footer_slot",
+        """
+        {
+          op: PLUGIN_OPERATIONS.Insert,
+          widget: {
+            id: 'custom_footer',
+            type: DIRECT_PLUGIN,
+            RenderWidget: () => (
+              <h1>This is the footer.</h1>
+            ),
+          },
+        }"""
+    )
+
+That's it!  If you rebuild the ``mfe`` image after enabling the plugin (via ``tutor images build mfe`` or ``tutor local launch``), "This is the footer." should appear at the bottom of every MFE.
+
+It's also possible to target a specific MFE's footer. For instance:
+
+.. code-block:: python
+
+    PLUGIN_SLOTS.add_items([
+        # Hide the custom footer
+        (
+            "profile",
+            "footer_slot",
+            """
+            {
+              op: PLUGIN_OPERATIONS.Hide,
+              widgetId: 'custom_footer',
+            }"""
+        ),
+        # Insert a footer just for the Profile MFE
+        (
+            "profile",
+            "footer_slot",
+            """
+            {
+              op: PLUGIN_OPERATIONS.Insert,
+              widget: {
+                id: 'custom_profile_footer',
+                type: DIRECT_PLUGIN,
+                RenderWidget: () => (
+                  <h1>This is the Profile MFE's footer.</h1>
+                ),
+              },
+            }"""
+        )
+    ])
+
+Note that here we're assuming you didn't remove the global footer configuration defined by the filter items targeting ``"all"``, so you have to hide ``custom_footer`` instead of ``default_contents``.  If you were to rebuild the MFE image now, the Profile MFE's footer would say "This is the Profile MFE's footer", whereas all the others would still contain the global "This is the footer." message.
+
+For more complex frontend plugins, you should make use of ``mfe-env-config-*`` patches to define your JSX components separately.  For instance, you could create an NPM plugin package, install it via ``mfe-dockerfile-post-npm-install``, import the desired components via ``mfe-env-config-buildtime-imports``, then refer to them with the ``PLUGIN_SLOTS`` filter as described above.  Refer to the `patch catalog <#template-patch-catalog>`_ below for more details.
+
+
 Installing from a private npm registry
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -410,6 +535,58 @@ This is the list of all patches used across tutor-mfe (outside of any plugin). A
     git clone https://github.com/overhangio/tutor-mfe
     cd tutor-mfe
     git grep "{{ patch" -- tutormfe/templates
+
+mfe-env-config-buildtime-imports
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Use this patch for any static imports you need in ``env.config.jsx``. They will be available here if you used the `mfe-docker-post-npm-install patch <#mfe-docker-post-npm-install>`_ to install an NPM package for all MFEs.
+
+It gets rendered at the very top of the file. You should use normal `ES6 import syntax <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import>`_.
+
+Note that if you want to only import a module for a particular MFE, doing it here won't work: you'll probably want to use the ``mfe-env-config-runtime-definitions-{}`` patch described below.
+
+File changed: ``tutormfe/templates/mfe/build/mfe/env.config.jsx``
+
+mfe-env-config-buildtime-definitions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Use this patch for arbitrary ``env.config.jsx`` javascript code that gets evaluated at build time. It is particularly useful for defining slightly more complex components for use in plugin slots.
+
+There's no version of this patch that runs per MFE.  If you want to define MFE-specific code, you should use the MFE-specific ``mfe-env-config-runtime-definitions-{}`` to achieve the same effect.
+
+File changed: ``tutormfe/templates/mfe/build/mfe/env.config.jsx``
+
+mfe-env-config-runtime-definitions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This patch gets rendered inside an ``async`` function in ``env.config.jsx`` that runs in the browser, allowing you to define conditional imports for external modules that may only be available at runtime. Just make sure to use `import() function <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import>`_ syntax:
+
+.. code-block:: javascript
+
+    const mymodule1 = await import('mymodule1');
+    const { default: myComponent } = await import('mymodule2');
+
+Note the second line in the example above: default module exports work a little differently with ``import()``.  To use the default export you can destructure the imported module, but you have to explicitly rename the ``default`` key, as `documented in MDN <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#importing_defaults>`_.
+
+Warning: if the dynamic import of a module fails for whatever reason, ``env.config.jsx`` execution will fail silently.
+
+File changed: ``tutormfe/templates/mfe/build/mfe/env.config.jsx``
+
+mfe-env-config-runtime-definitions-{}
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+With this patch you can conditionally import modules or define code for specific MFEs in ``env.config.jsx``. This is a useful place to put an import if you're using the ``mfe-docker-post-npm-install-*`` patch to install a plugin that only works on a particular MFE.
+
+As above, make sure to use the ``import()`` function.
+
+File changed: ``tutormfe/templates/mfe/build/mfe/env.config.jsx``
+
+mfe-env-config-runtime-final
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+At this point, ``env.config.jsx`` is ready to return the ``config`` object to the initialization code at runtime. You can use this patch to do anything to the object, including using modules that were imported dynamically earlier.
+
+File changed: ``tutormfe/templates/mfe/build/mfe/env.config.jsx``
 
 mfe-lms-development-settings
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -13,7 +13,7 @@ from tutor.hooks import priorities
 from tutor.types import Config, get_typed
 
 from .__about__ import __version__
-from .hooks import MFE_APPS, MFE_ATTRS_TYPE
+from .hooks import MFE_APPS, MFE_ATTRS_TYPE, PLUGIN_SLOTS
 
 # Handle version suffix in main mode, just like tutor core
 if __version_suffix__:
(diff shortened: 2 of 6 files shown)


arbrandes:
@regisb, how about this?

README.rst
```diff
@@ -308,6 +308,112 @@ In case you need to run additional instructions just before the build step you c
...
+    from tutormfe.hooks import PLUGIN_SLOTS
+
+    PLUGIN_SLOTS.add_item((
+        "all", "footer_slot",
+        """
+          {
+            /* Hide the default footer. */
+            op: PLUGIN_OPERATIONS.Hide,
+            widgetId: 'default_contents',
+          },
```
regisb, line 327: This is a bit of a nitpick: I suggest having separate items for every plugin slot. I.e:

```
 PLUGIN_SLOTS.add_items(
    ("all", "footer_slot",
        """
          {
            /* Hide the default footer. */
            op: PLUGIN_OPERATIONS.Hide,
            widgetId: 'default_contents'
          }),
    ("all", "footer_slot", """..."""),
    ...
)
```

In particular, with this approach we don't have to be too careful with trailing commas.

This is a minor concern and should not be blocking this PR.

tutormfe/plugin.py
```diff
@@ -90,6 +89,14 @@ def get_mfes() -> dict[str, MFE_ATTRS_TYPE]:
     return MFE_APPS.apply({})
 
 
+@functools.lru_cache(maxsize=None)
```
regisb, line 92: Make sure to use [tutor.hooks.lru_cache](https://docs.tutor.edly.io/reference/api/hooks/index.html#tutor.hooks.lru_cache) (and fix get_mfes above)

tutormfe/plugin.py
```diff
@@ -90,6 +89,14 @@ def get_mfes() -> dict[str, MFE_ATTRS_TYPE]:
...
+def get_plugin_slots(mfe_name: str) -> list[tuple[str, str, str]]:
+    """
+    This function is cached for performance.
+    """
+    return [i for i in PLUGIN_SLOTS.apply([]) if i[0] == mfe_name]
```
regisb, line 97: `PLUGIN_SLOTS.iterate()` is syntactic sugar for `PLUGIN_SLOTS.apply([])`.

tutormfe/templates/mfe/build/mfe/env.config.jsx
```diff
@@ -0,0 +1,48 @@
+{{- patch("mfe-env-config-static-imports") }}
+
+{{- patch("mfe-env-config-head") }}
```
regisb, line 3: Having two patch statements in quick succession is a clue that we should remove one. If some patches need to be at the bottom, then they can be added with a lower priority.

tutormfe/plugin.py
```diff
@@ -107,6 +114,15 @@ def iter_mfes() -> t.Iterable[tuple[str, MFE_ATTRS_TYPE]]:
     yield from get_mfes().items()
 
 
+def iter_plugin_slots(mfe_name: str) -> t.Iterable[tuple[str, str, str]]:
+    """
+    Yield:
+
+        (mfe_name, slot_name, slot_config)
```
regisb, line 121: We are never using `mfe_name` in the templates. Does that mean that we should simply not return it?

tutormfe/templates/mfe/build/mfe/env.config.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+      {%- for _, slot_name, slot_config in iter_plugin_slots(app_name) %}
+      config.pluginSlots.{{ slot_name }}.plugins.push(
+        {{- slot_config }}
+      );
+      {%- endfor %}
+    }
+    {%- endfor %}
+
+    {{- patch("mfe-env-config-tail") }}
```
regisb, line 42: nitpick: The "-tail" and "-head" suffixes are not used for other patch statements in Tutor. For instance, the patches in openedx/Dockerfile use "-pre", "-post", "-final". I'm not sure which convention is best, but we should probably follow the existing one.

regisb (commented):
I think this is a great improvement, and I only have minor comments. Looking forward to merging this.

tutormfe/plugin.py
arbrandes: Got it, will do!

tutormfe/templates/mfe/build/mfe/env.config.jsx
regisb: OK let's keep those patches then.

arbrandes:
@regisb, ready for a second round.

tutormfe/templates/mfe/build/mfe/env.config.jsx
```diff
@@ -0,0 +1,51 @@
+{{- patch("mfe-env-config-buildtime-imports") }}
+
+function addPlugins(config, slot_name, plugins) {
+  if (slot_name in config.pluginSlots === false) {
```
regisb, line 4: Is this idiomatic? Shouldn't it be `if(!(slot_name in config.pluginSlots))`? Or `if(config.pluginSlots[slot_name] === null)`?

regisb (approved):
Stellar!

tutormfe/templates/mfe/build/mfe/env.config.jsx
```diff
@@ -0,0 +1,51 @@
...
+     * MFEs, so we import it dynamically. In addition, for dynamic imports to
+     * work with Webpack all of the code that actually uses the imported module
+     * needs to be inside the `try{}` block.
+     */
+    const { DIRECT_PLUGIN, PLUGIN_OPERATIONS } = await import('@openedx/frontend-plugin-framework');
+
+    {{- patch("mfe-env-config-runtime-definitions") }}
+
+    {%- for slot_name, plugin_config in iter_plugin_slots("all") %}
+    addPlugins(config, '{{ slot_name }}', [{{ plugin_config }}]);
```
hinakhadim, line 32: Is there any particular reason for adding plugin_config object one by one?

Could we pass array for one plugin_slot (keeping in mind multiple plugins can have same slots)?
image

Like this:
```
PLUGIN_SLOTS.add_items([
        (
            mfe,
            "footer_slot",
            """
            [
    {
        op: PLUGIN_OPERATIONS.Hide,
        widgetId: 'default_contents',
    }
    {
      op: PLUGIN_OPERATIONS.Insert,
      widget: {
        id: 'default_contents',
        type: DIRECT_PLUGIN,
        priority: 1,
        RenderWidget: <Footer />,
      },
    },
    {
      op: PLUGIN_OPERATIONS.Insert,
      widget: {
        id: 'read_theme_cookie',
        type: DIRECT_PLUGIN,
        priority: 2,
        RenderWidget: AddDarkTheme,
      },
    },
  ]"""
        ),
    ])
```

tutormfe/templates/mfe/build/mfe/env.config.jsx
hinakhadim: I think only `if (slot_name in config.pluginSlots)` is fine.

tutormfe/templates/mfe/build/mfe/env.config.jsx
hinakhadim: Apologies, I just realized that objects are already passing in array. I need to pass without array brackets:
```javascript

PLUGIN_SLOTS.add_items([
        (
            mfe,
            "footer_slot",
            """
            
    {
        op: PLUGIN_OPERATIONS.Hide,
        widgetId: 'default_contents',
    },
    {
      op: PLUGIN_OPERATIONS.Insert,
      widget: {
        id: 'default_contents',
        type: DIRECT_PLUGIN,
        priority: 1,
        RenderWidget: <Footer />,
      },
    },
    {
      op: PLUGIN_OPERATIONS.Insert,
      widget: {
        id: 'read_theme_cookie',
        type: DIRECT_PLUGIN,
        priority: 2,
        RenderWidget: AddDarkTheme,
      },
    },
  """
        ),
    ])
```

README.rst
```diff
@@ -308,6 +308,131 @@ In case you need to run additional instructions just before the build step you c
...
+            "all",
+            "footer_slot",
+            """
+            {
+              op: PLUGIN_OPERATIONS.Hide,
+              widgetId: 'default_contents',
+            }"""
+        ),
+
+The last parameter to ``add_item()`` is a big string with the actual slot configuration, which will be interpreted as JSX. What we're doing there is hiding the default contents of the footer with a ``PLUGIN_OPERATIONS.Hide``. (You can refer to the `frontend-plugin-framework README <https://github.com/openedx/frontend-plugin-framework/#>`_ for a full description of the possible plugin types and operations.) And the ``default_contents`` widget ID we're targetting always refers to what's in an unconfigured slot by default.
```
hinakhadim, line 373: nit: targetting -> targeting

README.rst
```diff
@@ -411,6 +536,56 @@ This is the list of all patches used across tutor-mfe (outside of any plugin). A
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This patch gets rendered inside an ``async`` function in ``env.config.jsx`` that runs in the browser, allowing you to define conditional imports for external modules that may only be available at runtime. Just make sure to use `import() function <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import>`_ syntax:
+
+.. code-block:: javascript
+
+    const mymodule1 = await import('mymodule1');
+    const { mycomponent1, mycomponent2 } = await import('mymodule2');
+
+Note that if any module is not available at runtime, ``env.config.jsx`` execution will fail silently.
```
hinakhadim, line 569: Could we update the documentation to specify that if a component is exported as a default export, it should be accessed using Component.default when dynamically imported? This clarification will help ensure proper usage and avoid potential errors during runtime.

I was testing with indigo and here's my file:
```
// import Footer from '@edly-io/indigo-frontend-component-footer';

...

if (process.env.APP_ID == 'learner-dashboard') {
const Footer = await  import('@edly-io/indigo-frontend-component-footer');
...
{
      op: PLUGIN_OPERATIONS.Insert,
      widget: {
        id: 'default_contents',
        type: DIRECT_PLUGIN,
        priority: 1,
        RenderWidget: <Footer />,
      },
    },
...
}

```

It is throwing error `Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.` 

If I use `const { Footer } = await  import('@edly-io/indigo-frontend-component-footer');`, it throws this error `Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.`.

The footer is exported as default. `export default Footer;`. The errors are generated when I shift to dynamic import `await (...)`. Could we add in doc that we have to use `Footer(Component).default` in case of dynamic imports.

hinakhadim (approved):

tutormfe/templates/mfe/build/mfe/env.config.jsx
arbrandes: On syntax: I guess it's a matter of opinion.  MDN [likes `=== false`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in#try_it) for negating the `in` operator.  But I do thing `in` is the best check, here.  It's unambiguous as to whether the property is defined or not.

But we do have to use the negation, here.  It avoids duplicating code.

tutormfe/templates/mfe/build/mfe/env.config.jsx
```diff
@@ -0,0 +1,51 @@
...
+      plugins: []
+    };
+  }
+
+  config.pluginSlots[slot_name].plugins.push(...plugins);
+}
+
+{{- patch("mfe-env-config-buildtime-definitions") }}
+
+async function getConfig () {
```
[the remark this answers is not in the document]
arbrandes: Sure.  But I don't know what the behavior is going to be if you actually call frontend-platform's getConfig while actually setting config.

In any case, I'll just call it `setConfig`, then.

tutormfe/templates/mfe/build/mfe/env.config.jsx
arbrandes: There are different ways you can load the configuration in.  @regisb suggested we document it one-by-one to avoid issues with trailing commas.  Though in this latest iteration of `env.config.jsx` it doesn't really matter if there are trailing commas or not, I guess it keeps it simple.

Otherwise, yes you can totally use a single `add_item()` with a string that has multiple ops for one slot.

README.rst
arbrandes: Sure, I'll point to the MDN documentation.  They mention that you have to use [the `default` key](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#module_namespace_object).

arbrandes: They actually document a better way [further down in the page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#importing_defaults), which is what I ended up putting up as an example.

regisb:
@DawoudSheraz can you please merge and rebase the sumac branch on top of nightly?

DawoudSheraz:
> I will test this branch locally with indigo nightly. Hina is on leaves this week, she mentioned that this can break indigo. I would verify that before merging.

I tested this locally and indigo's dark theme toggle is not working as expected. It is not appearing consistently across pages (for instance, it does not appear on the learner dashboard). Even when it appears, the toggle does not retain its value when navigating to different pages. Hina mentioned that this could impact indigo. 

I am ok with merging considering Hina will be back from leaves next week and can look into updating Indigo.

arbrandes:
Let me know if I can help with anything.  We should probably try and get this (or something like this) merged this week so it comes out with Sumac.

regisb:
@hinakhadim friendly nudge :)

hinakhadim:
Hi @arbrandes ,

There's an issue generating with this PR on tutor-indigo that is it's not picking up by any MFE. The image building is successful and when we try to see in browser. The changes are not appearing. Steps:
- install indigo footer in leaner-dashboard MFE
- the contents of `env.config.jsx` file:
```
import React, { useEffect } from 'react';
import Cookies from 'universal-cookie';

import { getConfig } from '@edx/frontend-platform';

function addPlugins(config, slot_name, plugins) {
  // ...
}

let themeCookie = 'indigo-toggle-dark';
let themeCookieExpiry = 90; // days

const AddDarkTheme = () => {
  // ...

  return (<div />);
};

async function setConfig () {
  let config = {
    pluginSlots: {}
  };

  try {
    /* We can't assume FPF exists, as it's not declared as a dependency in all
     * MFEs, so we import it dynamically. In addition, for dynamic imports to
     * work with Webpack all of the code that actually uses the imported module
     * needs to be inside the `try{}` block.
     */
    const { DIRECT_PLUGIN, PLUGIN_OPERATIONS } = await import('@openedx/frontend-plugin-framework');

const Footer = await import(/* webpackIgnore: true */`@edly-io/indigo-frontend-component-footer`).catch(() => null); // Fallback to null if not available

    if (process.env.APP_ID == 'authn') {
    }
    if (process.env.APP_ID == 'authoring') {
    }
    
    if (process.env.APP_ID == 'learner-dashboard') {
      addPlugins(config, 'footer_slot', [ 
            {
                op: PLUGIN_OPERATIONS.Hide,
                widgetId: 'default_contents',
            },
            {
                op: PLUGIN_OPERATIONS.Insert,
                widget: {
                    id: 'default_contents',
                    type: DIRECT_PLUGIN,
                    priority: 1,
                    RenderWidget: Footer.default,
                },
            },
            {
            op: PLUGIN_OPERATIONS.Insert,
                widget: {
                    id: 'read_theme_cookie',
                    type: DIRECT_PLUGIN,
                    priority: 2,
                    RenderWidget: AddDarkTheme,
                },
            },
  ]);
    }
   
  } catch { }

  return config;
}

export default setConfig;
```

- I was thinking that `AddDarkTheme` func is not working. But the default footer is not hidden and new Indigo footer is not applied. 
- I added the previous `env.config.jsx` that exists in Indigo. and that works perfectly.


Reason of Issue: 
I'm assuming that this is causing because of `setConfig`. `export default setConfig` is a function and in the previous `env.config.jsx` (already used in indigo now), we are exporting object `let config = {...}; export default config;`


Could you please help in resolving it? I tried to figure out by calling `setConfig` but it didn't work. Because it's also `async` function and return Promise. Also, if you help how to add logger to `env.config.jsx`, then it helps in debugging it easily.

arbrandes:
@hinakhadim, lemme try and reproduce it on my end.  Is there a PR to tutor-indigo that I can refer to?

hinakhadim:
Sure, please. Here's the PR Link https://github.com/overhangio/tutor-indigo/pull/109

It might be `webpack Ignore` is causing this issue because I remember it was working before.  But the problem is `ora-grading` fails the build if I don't import Footer with `webpack Ignore`. (FYI: I also tried by importing Footer in `if` conditions. And one thing is that we aren't included footer in ora-grading but still I don't have idea why it's reading the Footer import during build and produces error). 

Error is in this conversation: https://github.com/overhangio/tutor-mfe/pull/240#issuecomment-2511793463 . At that time my env.config.jsx was like: (As you can see, footer in even imported in `learner-dashboard condition`. It has nothing with `ora-grading, but still failed the build`.
```
...
try{

const { DIRECT_PLUGIN, PLUGIN_OPERATIONS } = await import('@openedx/frontend-plugin-framework');
 
     if (process.env.APP_ID == 'authn') {
    }
    ...
    if (process.env.APP_ID == 'learner-dashboard') {
        const Footer = await import("@edly-io/indigo-frontend-component-footer");
         
        addPluginSlot(...);
    }
   ...
   if (process.env.APP_ID == 'ora-grading') {
    }
    ...
} catch {}
...
```


After adding `webpack Ignore` , image building is becoming successful but MFEs are not picking this file or may be picking but not applying any plugin slot.

arbrandes:
@hinakhadim, I have some findings to discuss, but I'll do it as a review to https://github.com/overhangio/tutor-indigo/pull/109.

arbrandes:
@hinakhadim, now that the MFEs have been fixed, let me know if things are good with https://github.com/overhangio/tutor-indigo/pull/109 so we can merge this.

hinakhadim:
Things are working good. Just waiting for a heads-up from another member regarding testing before we proceed with the merge. Status: https://github.com/overhangio/tutor-indigo/pull/109#issuecomment-2523864661

DawoudSheraz:
@arbrandes Hi. Please fix the failing tests so that this can be merged. Thanks

arbrandes:
@DawoudSheraz, I'm looking into it, but to be honest, it'll probably take me a while.  My expertise with Python type annotations is limited.

arbrandes:
Plus, I'm not currently set up to reproduce the test environment. 😓

tutormfe/plugin.py
```diff
@@ -82,20 +81,21 @@ def _add_core_mfe_apps(apps: dict[str, MFE_ATTRS_TYPE]) -> dict[str, MFE_ATTRS_T
...
-@tutor_hooks.Actions.PLUGIN_LOADED.add()
-def _clear_get_mfes_cache(_name: str) -> None:
+@tutor_hooks.lru_cache
+def get_plugin_slots(mfe_name: str) -> list[tuple[str, str]]:
     """
-    Don't forget to clear cache, or we'll have some strange surprises...
+    This function is cached for performance.
     """
-    get_mfes.cache_clear()
+    i: tuple[str, str, str]
```
regisb, line 97: This annotation should not be necessary, let's get rid of it.

hinakhadim:
For the issue at line 124, I'm not entirely sure if this is the most appropriate solution, but it does resolve the problem
```
def get_mfe(mfe_name: str) -> MFE_ATTRS_TYPE:
    mfes : dict[str, MFE_ATTRS_TYPE] = get_mfes() 
    return mfes.get(mfe_name, {})
```

tutormfe/plugin.py
arbrandes: Yeah, this was just a dumb attempt to get the linting to stop complaining before I got python 3.9 set up locally. 😅

tutormfe/plugin.py
```diff
@@ -82,20 +81,21 @@ def _add_core_mfe_apps(apps: dict[str, MFE_ATTRS_TYPE]) -> dict[str, MFE_ATTRS_T
...
+    return [i[-2:] for i in PLUGIN_SLOTS.iterate() if i[0] == mfe_name]
```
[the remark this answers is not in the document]
arbrandes: Nice catch!  This solved the linting cascade.

tutormfe/plugin.py
```diff
@@ -107,11 +106,20 @@ def iter_mfes() -> t.Iterable[tuple[str, MFE_ATTRS_TYPE]]:
...
+    """
+    yield from get_plugin_slots(mfe_name)
+
+
 def is_mfe_enabled(mfe_name: str) -> bool:
     return mfe_name in get_mfes()
 
 
-def get_mfe(mfe_name: str) -> MFE_ATTRS_TYPE:
+def get_mfe(mfe_name: str) -> t.Union[MFE_ATTRS_TYPE, t.Any]:
```
arbrandes, line 122: @hinakhadim, your previous comment got me on the right track, thanks - but I didn't want to change running code just for the sake of a type hint, so this is what I did.  It seems there's significant precedent for `Any` in the codebase, so I expect it to be fine.

arbrandes:
Ok, finally got pyenv properly set up so I could reproduce the failures.  Thanks for the reviews!

tutormfe/plugin.py
regisb: The issue is caused by the lru_cache decorator, which is not correctly typed. You can bypass this issue without `Any` by writing:

```
def get_mfe(mfe_name: str) -> MFE_ATTRS_TYPE:
    mfe: MFE_ATTRS_TYPE = get_mfes().get(mfe_name, {})
    return mfe
```

PR #241, opened 2024-12-14
fix: render mfe-ports for  unmounted mfes when these exist

hinakhadim:
### Description
The issue occurs when all MFEs (Micro Frontends) are mounted, and the `ports` section in the Docker configuration is left empty, leading to a syntax error in the generated Dockerfile. Specifically, the `ports:` line becomes invalid as another key follows it (`auth:`) without any content. In the `tutor dev` context, the issue occurs because developers mount MFEs for testing purposes. It won't reproduce for `tutor local`.

```
...

mfe:
    ports:
auth:
   ....
```

### How to Reproduce
- Clone all MFEs (`authn, account, communications, course-authoring, discussions, gradebook, learner-dashboard, learning, ora-grading, profile`)
- Run `tutor mounts add $(pwd)/frontend-app-*` where `*` represents all MFEs name one by one
- `tutor config save`
- `tutor dev start` OR directly check the `mfe-ports` section in `tutor-env/env/dev/docker-compose.yml` 

### What this PR does
This PR resolves the issue by implementing logic to first check if any unmounted MFEs exist. If unmounted MFEs are found, their corresponding ports are rendered in the ports: section. This ensures that the Docker configuration is generated correctly and avoids errors caused by leaving the `ports:` section empty.

### TestCases:
- [x] If any unmounted MFE exists, its corresponding port should be rendered under the `mfe: ports:` section.
```
...

mfe:
    ports:
        - 1997:8002 # account
        - 1984:8002 # communications

authn:
    image: ...
...           
```
- [x] If no unmounted MFEs exist, the `mfe: ports:` section should be excluded entirely from the generated output.
```
...
#There should be no section of mfe->ports

authn:
     image: ....
...          
```

Closes https://github.com/overhangio/tutor-mfe/issues/242

### DIFF
--- a/tutormfe/patches/local-docker-compose-dev-services
+++ b/tutormfe/patches/local-docker-compose-dev-services
@@ -1,10 +1,17 @@
+{%- set unmounted_mfes = [] %}
+{%- for app_name, app in iter_mfes() %}
+   {%- if not iter_mounts(MOUNTS, app_name)|list %}
+       {%- set _ = unmounted_mfes.append((app_name, app)) %}
+   {%- endif %}
+{%- endfor %}
+{% if unmounted_mfes|length > 0 %}
 mfe:
-    ports:
-        {%- for app_name, app in iter_mfes() %}
-        {%- if not iter_mounts(MOUNTS, app_name)|list %}
-        - {{ app["port"] }}:8002 # {{ app_name }}
-        {%- endif %}
-        {%- endfor %}
+   ports:
+       {%- for app_name, app in unmounted_mfes %}
+       - {{ app["port"] }}:8002 # {{ app_name }}
+       {%- endfor %}
+{% endif %}
+
 
 {%- for app_name, app in iter_mfes() %}
 {%- set mounts = iter_mounts(MOUNTS, app_name)|list %}
--- a/tutormfe/patches/local-docker-compose-dev-services
+++ b/tutormfe/patches/local-docker-compose-dev-services
@@ -1,8 +1,6 @@
-{%- set unmounted_mfes = [] %}
+{%- set mfe_data = MFEMountData(MOUNTS) %}
 
-{%- for app_name, app in iter_mfes() %}
-{%- set mounts = iter_mounts(MOUNTS, app_name)|list %}
-{%- if mounts %}
+{%- for app_name, app, mounts in mfe_data.mounted %}
 {{ app_name }}: # Work on this MFE for development
     image: "{{ MFE_DOCKER_IMAGE_DEV_PREFIX }}-{{ app_name }}-dev:{{ MFE_VERSION }}"
     ports:
@@ -19,16 +17,13 @@
         - lms
     environment:
         - "PORT={{ app['port'] }}"
-{%- else %}
-{%- set unmounted_mfes = unmounted_mfes.append((app_name, app)) %}
-{%- endif %}
 {%- endfor %}
 
 
-{% if unmounted_mfes|length > 0 %}
+{% if mfe_data.unmounted|length > 0 %}
 mfe:
    ports:
-       {%- for app_name, app in unmounted_mfes %}
+       {%- for app_name, app in mfe_data.unmounted %}
        - {{ app["port"] }}:8002 # {{ app_name }}
        {%- endfor %}
 {% endif %}
--- a/tutormfe/patches/local-docker-compose-dev-services
+++ b/tutormfe/patches/local-docker-compose-dev-services
@@ -1,17 +1,4 @@
 {%- set unmounted_mfes = [] %}
-{%- for app_name, app in iter_mfes() %}
-   {%- if not iter_mounts(MOUNTS, app_name)|list %}
-       {%- set _ = unmounted_mfes.append((app_name, app)) %}
-   {%- endif %}
-{%- endfor %}
-{% if unmounted_mfes|length > 0 %}
-mfe:
-   ports:
-       {%- for app_name, app in unmounted_mfes %}
-       - {{ app["port"] }}:8002 # {{ app_name }}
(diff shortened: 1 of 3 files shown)


tutormfe/patches/local-docker-compose-dev-services
```diff
@@ -1,10 +1,17 @@
+{%- set unmounted_mfes = [] %}
+{%- for app_name, app in iter_mfes() %}
+   {%- if not iter_mounts(MOUNTS, app_name)|list %}
+       {%- set _ = unmounted_mfes.append((app_name, app)) %}
+   {%- endif %}
+{%- endfor %}
+{% if unmounted_mfes|length > 0 %}
```
regisb, line 7: nit: `{% if unmounted_mfes|length > 0 %}`

tutormfe/patches/local-docker-compose-dev-services
```diff
@@ -1,10 +1,17 @@
+{%- set unmounted_mfes = [] %}
+{%- for app_name, app in iter_mfes() %}
```
regisb, line 2: Could we simplify this declaration with the following?

```
{%- set unmounted_mfes = [(app_name, app) for app_name, app in iter_mfes() if not list(iter_mounts(MOUNTS, app_name))] %}
```

regisb (changes requested):
This solution is way better than the one I initially suggested. Please just add a changelog entry and add "close #242" to your commit message.

tutormfe/patches/local-docker-compose-dev-services
```diff
@@ -1,10 +1,17 @@
...
+   {%- if not iter_mounts(MOUNTS, app_name)|list %}
+       {%- set _ = unmounted_mfes.append((app_name, app)) %}
```
DawoudSheraz, line 4: shouldn't append be enough by itself? Is it necessary to set the output to _?

tutormfe/patches/local-docker-compose-dev-services
hinakhadim: I've tried to applied using the below code.
```python
def get_unmounted_mfes() -> list:
    unmounted_mfes = []
    def append_to_unmounted_mfes(mounts: list[str], app_name:str, app: MFE_ATTRS_TYPE):
        if not mounts:
            unmounted_mfes.append((app_name, app))

        return unmounted_mfes
    return append_to_unmounted_mfes


tutor_hooks.Filters.ENV_TEMPLATE_VARIABLES.add_items(
    [
        ...
        ("get_unmounted_mfes", get_unmounted_mfes()),
    ]
)
```

That leads us to this template:
```
{%- set unmounted_mfes = [] %}

{%- for app_name, app in iter_mfes() %}
{%- set mounts = iter_mounts(MOUNTS, app_name)|list %}
{%- set unmounted_mfes = get_unmounted_mfes(mounts, app_name, app) %}  ---> here I add this line
{%- if mounts %}
{{ app_name }}: # Work on this MFE for development
...
{%- endif %}
{%- endfor %}



{% if unmounted_mfes|length > 0 %}
mfe:
   ports:
       {%- for app_name, app in unmounted_mfes %}
       - {{ app["port"] }}:8002 # {{ app_name }}
       {%- endfor %}
{% endif %}
```

tutormfe/patches/local-docker-compose-dev-services
hinakhadim: The `set _` is necessary. Jinja does not support direct in-place modifications like Python does (list.append()), so you need to reassign the list with the new value.
https://stackoverflow.com/a/66019113

tutormfe/patches/local-docker-compose-dev-services
hinakhadim: I have re-arranged the code in the patch.

regisb: I'm not a big fan of this pattern, as it moves too much declarative code to the Jinja templates. This is not a great practice. Instead, I suggest:

```
def get_unmounted_mfes(mounts):
    unmounted_mfes = []   
     for mfe in ...:
        if ...:
            unmounted_mfes.append(mfe)
    return unmounted_mfes

tutor_hooks.Filters.ENV_TEMPLATE_VARIABLES.add_item(
        ("get_unmounted_mfes", get_unmounted_mfes),
)
```

and then in the templates:

    {%- set unmounted_mfes = get_unmounted_mfes(MOUNTS) %}

regisb: @hinakhadim did you consider my suggestion above?

hinakhadim: I missed this message. I'll try with the suggested one. Thanks

tutormfe/patches/local-docker-compose-dev-services
```diff
@@ -25,5 +19,16 @@ mfe:
...
 {%- endfor %}
+
+
+{% if unmounted_mfes|length > 0 %}
+mfe:
+   ports:
+       {%- for app_name, app in unmounted_mfes %}
+       - {{ app["port"] }}:8002 # {{ app_name }}
+       {%- endfor %}
+{% endif %}
```
DawoudSheraz, line 34: nit: missing end of file line.

DawoudSheraz (commented):
Please add a changelog entry

DawoudSheraz (approved):
I did not test it locally but the code looks good. Nice re-organization @ahmed-arb

DawoudSheraz:
@CodeWithEmad FYI as this fixes https://github.com/overhangio/tutor-mfe/issues/242.

CodeWithEmad:
Great work, folks 🤘🏼

arbrandes (approved):
Looks great, thanks!

PR #244, opened 2025-02-21
build: migrate to hatch/pyproject.toml

DawoudSheraz:
migrate from setup.py.setuptools to pyproject.toml/hatch, completes tutor-mfe action item from https://github.com/overhangio/tutor/issues/1190


pyproject.toml
```diff
@@ -1,2 +1,70 @@
+# https://packaging.python.org/en/latest/tutorials/packaging-projects/
+# https://hatch.pypa.io/latest/config/build/
+
+[project]
+name = "tutor-mfe"
+license = {text = "AGPLv3" }
```
Danyal-Faheem, line 6: We should use the licenses as defined in the [SPDX license list](https://spdx.org/licenses/) and keep them consistent across the plugins.

Let's change this to `AGPL-3.0-only` as defined in other plugins as well.

pyproject.toml
```diff
@@ -1,2 +1,70 @@
...
+build-backend = "hatchling.build"
+
+[tool.hatch.build.targets.sdist]
+# Disable strict naming, otherwise twine is not able to detect name/version
+strict-naming = false
+include = ["/tutormfe"]
+exclude = ["tests*"]
+
+[tool.hatch.build.targets.wheel]
+packages = ["tutorminio"]
```
Danyal-Faheem, line 70: Typo here, this should be `tutormfe` instead.

pyproject.toml
```diff
@@ -1,2 +1,70 @@
...
+mfe = "tutormfe.plugin"
+
+# https://packaging.python.org/en/latest/specifications/well-known-project-urls/#well-known-labels
+[project.urls]
+Homepage = "https://docs.tutor.edly.io/"
+Documentation = "https://docs.tutor.edly.io/"
+Code = "https://github.com/overhangio/tutor-mfe"
+Issues = "https://github.com/overhangio/tutor-mfe/issues"
+Changelog = "https://github.com/overhangio/tutor-mfe/blob/release/CHANGELOG.md"
+Community = "https://discuss.openedx.org/tag/tutor"
```
Danyal-Faheem, line 53: The `tutor-mfe` [tag ](https://discuss.openedx.org/tag/tutor-mfe) is also a widely used tag on the discussion forum. Maybe we should use that here?

Danyal-Faheem (changes requested):
Looks good, just needs some minor revisions

pyproject.toml
DawoudSheraz: Ah, I missed it. I was updating another PR and forgot I had to do this in here. 
One thing: this is showing up in pip show after installation. The docs suggest/recommend using the standard but there does not seem to be an enforcement.

pyproject.toml
DawoudSheraz: tutor-mfe has fewer posts compared to tutor. We can use both as separate URLs, I think.

Danyal-Faheem (approved):
LGTM! Everything seems to be working for me

arbrandes (approved):
Just read up on the parent issue, and have no objections.  👍🏼

PR #246, opened 2025-03-24
chore: remove unnecessary settings and waffle flags for account and profile MFEs

Danyal-Faheem:
Fixes part of https://github.com/overhangio/tutor/issues/1207.

As the site-configuration script is planned for deprecation in tutor-core, we look to remove or update its usage with a suitable replacement.

However, as of https://github.com/openedx/edx-platform/pull/36219, we do not require the site-configuration entries for the profile MFE anymore as the MFE is the default frontend going forward.

Similarly, the `redirect_to_microfrontend` flags are also not required anymore for the account and profile MFE as they are now the default frontends going forward.


DawoudSheraz (approved):

arbrandes (approved):
Makes sense!

PR #249, opened 2025-04-01
feat: enable the new Studio Unit Page by default

arbrandes:
As of the merging of [1], the new unit page is ready for general use, so enable it by default.

[1] https://github.com/openedx/frontend-app-authoring/pull/1492


regisb (approved):
yup, all good!

PR #251, opened 2025-04-04
Add section on deploying MFE changes to production

sarina:
Per discussion in https://discuss.openedx.org/t/operation-of-the-mfe/15396/5

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -478,8 +478,6 @@ This works for custom MFEs, as well. For example, if you added your own MFE name
 
     tutor mounts add /path/to/frontend-app-myapp
 
-Similarly, in production, the "mfe" Docker image will be rebuilt automatically during ``tutor local launch``.
-
 .. note::
 
   Docker tries to run as many build processes in parallel as possible, but this can cause failures in the MFE image build.  If you're running into OOM issues, RAM starvation, or network failures during NPM installs, try the following before restarting::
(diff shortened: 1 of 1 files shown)


arbrandes (approved):
Looks great to me!

README.rst
```diff
@@ -490,6 +488,13 @@ Similarly, in production, the "mfe" Docker image will be rebuilt automatically d
     EOF
     docker buildx create --use --name=singlecpu --config=./buildkitd.toml
 
+Deploying Changes to Production
+-------------------------------
+
+You will need to rebuild the mfe Docker image with your changes, and then deploy it to production - exactly how depends on whether you are running ``tutor local`` or ``tutor k8s``.
```
DawoudSheraz, line 494: I think this section can be removed. The next section has the appropriate details (the codebase is on same machine, build image, and the restart). We can mention the production environment (k8s vs local) in that section.

sarina: @DawoudSheraz Per discussion in https://discuss.openedx.org/t/operation-of-the-mfe/15396/5 people are getting stuck here. I don't think removing this is the right call.

arbrandes: I believe this adds relevant information with more visibility, in a section with an anchor so we can refer people directly to it.  I agree with @sarina that this questions comes up often enough that it warrants making it more easily accessible.

DawoudSheraz: Oh, I did not mean removing the entire section. I meant the first paragraph (🤦🏽 ). The second paragraph includes the necessary details:
- The codebase is on the same machine
- Building the image
- Start after build

The only missing item is the selection of local vs k8s, which is mentioned in the first paragraph.  That can be added to 2nd one.

sarina: You're suggesting to remove this sentence (which provides context to people who already know what they're doing) and only have the second sentence? I disagree - I think the context is helpful here, and more explicit. Docs should be easy to read and follow, they do not need to be the most concise thing.

DawoudSheraz (approved):

sarina:
@DawoudSheraz or @arbrandes could one of you merge this for me?

PR #252, opened 2025-04-09
[DRAFT] feat! replace webpack by rsbuild for faster build

regisb:
This is the result of a long investigation to accelerate the building of MFE apps in Open edX. Many of the intermediate results can be found in this issue: https://github.com/openedx/wg-frontend/issues/184

The tl;dr is that building MFEs takes a very long time, and that's because of Webpack. In this PR, we replace Webpack by [rsbuild](https://rsbuild.dev/), resulting in a considerable performance improvement.

This PR is dependent on the following upstream PRs:

- https://github.com/openedx/frontend-app-communications/pull/223
- https://github.com/openedx/frontend-app-gradebook/pull/446
- https://github.com/openedx/frontend-app-ora-grading/pull/408
- https://github.com/openedx/frontend-app-learner-dashboard/pull/598

## Open questions

- **Does this still support env.config.jsx customisations?** I think it should, but I haven't checked.
- **Does this build Paragon?** I have no idea. Probably not. To this day I still don't know what's Paragon.

## TODO

Things we need to address before (if?) we merge:

- Remove references to regisb/rsbuild in plugin.py
- Pin rsbuild dependencies in Dockerfile
- Add changelog entry

### Configuration

These benchmarks are performed on my personal computer: a Intel(R) Core(TM) i5-4670K CPU @ 3.40GHz with 4 processors and 20 GB RAM. My custom Docker builder runs with `max-parallelism = 4`, as per [these instructions](https://docs.tutor.edly.io/troubleshooting.html#high-resource-consumption-by-docker-on-tutor-images-build):

```
$ docker buildx ls
NAME/NODE        DRIVER/ENDPOINT                           STATUS    BUILDKIT   PLATFORMS
max4cpu*         docker-container                                               
 \_ max4cpu0      \_ unix:///var/run/docker.sock           running   v0.13.1    linux/amd64 (+3), linux/386
```

All plugins are disabled, except for "mfe" and "forum", and there is no bind-mount:

```
tutor plugins disable all
tutor plugins enable mfe forum
tutor mounts list
```

The "mfe" image is built from scratch on the "main" branch; then, we re-build the image by clearing the cache for the "*-prod" build steps, which are the ones that run `npm run build`:

```
# Update environment
tutor config save
# Fill cache
tutor images build mfe
# Build without cache
time tutor images build --docker-arg=--no-cache-filter=account-prod,authn-prod,authoring-prod,communications-prod,discussions-prod,ecommerce-prod,gradebook-prod,learner-dashboard-prod,learning-prod,ora-grading-prod,payment-prod,profile-prod mfe
```

The changes in this PR are tested by checking out this branch:

```
cd <path to>/tutor-mfe
git checkout regisb/rsbuild
```

Then run the commands above.

We list the benchmarks by date, because we expect these figures to change as we iterate on this PR.

#### Reference 1: total build time

On the main branch, the total elapsed ("real") time of that last step is **468 s** (7 min 48 s). This will serve as a point of reference for all benchmarks.

#### Reference 2: no parallelism build of the learning MFE

Another interesting point of reference is the time that is needed to build the learning MFE, without parallelism -- that is, without concurrent builds of other MFEs. This is a figure that can be obtained by running:

```
time tutor images build --docker-arg=--no-cache-filter=learning-prod --target=learning-prod mfe
```

And then checking the time to build the MFE in the Docker logs: 

```
 => [learning-prod 1/1] RUN npm run build                                119.5s
```

Here, the reference time is **119.5s**.

### 2025/04/10 Locked cache

We added the following locked cache in the build step:

```
RUN --mount=type=cache,target=/openedx/app/node_modules/.cache,sharing=locked rsbuild build
```

This causes all builds to run sequentially, and not in parallel. The good thing with this change is that it allows us to no longer customise our Docker builder with `max-parallelism=x`. The only downside is that there might be a drop in performance; but in practice, we don't really observe a noticeable one.

- Ref 1: 96s
- Ref 2: 9.2s (practically unchanged)

### 2025/04/09 Initial results

- **Ref 1:** **85-105s** (4.46-5.5x improvement)
- **Ref 2:** **7.1s** (14.8x improvement)

All MFEs build under 23s, despite the fact that there are 4 builds in parallel:

```
 => [communications-prod 2/2] RUN rsbuild build                                       21.7s
 => [learner-dashboard-prod 2/2] RUN rsbuild build                                    23.0s
 => [learning-prod 2/2] RUN rsbuild build                                             22.7s
 => [discussions-prod 2/2] RUN rsbuild build                                          22.6s
 => [gradebook-prod 2/2] RUN rsbuild build                                            24.1s
 => [ora-grading-prod 2/2] RUN rsbuild build                                          22.6s
 => [account-prod 2/2] RUN rsbuild build                                              22.7s
 => [authn-prod 2/2] RUN rsbuild build                                                22.7s
 => [profile-prod 2/2] RUN rsbuild build                                              18.6s
 => [authoring-prod 2/2] RUN rsbuild build                                            18.5s
```

### DIFF
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -32,6 +32,7 @@ config = {
 }
 
 CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
+    # TODO restore vanilla repositories
     "authn": {
         "repository": "https://github.com/openedx/frontend-app-authn.git",
         "port": 1999,
@@ -45,7 +46,9 @@ CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
         "port": 1997,
     },
     "communications": {
-        "repository": "https://github.com/openedx/frontend-app-communications.git",
+        "repository": "https://github.com/regisb/frontend-app-communications.git",
+        "version": "regisb/rsbuild",
+        # "repository": "https://github.com/openedx/frontend-app-communications.git",
         "port": 1984,
     },
     "discussions": {
@@ -53,11 +56,15 @@ CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
         "port": 2002,
     },
     "gradebook": {
-        "repository": "https://github.com/openedx/frontend-app-gradebook.git",
+        "repository": "https://github.com/regisb/frontend-app-gradebook.git",
+        "version": "regisb/rsbuild",
+        # "repository": "https://github.com/openedx/frontend-app-gradebook.git",
         "port": 1994,
     },
     "learner-dashboard": {
-        "repository": "https://github.com/openedx/frontend-app-learner-dashboard.git",
+        "repository": "https://github.com/regisb/frontend-app-learner-dashboard.git",
+        "version": "regisb/rsbuild",
+        # "repository": "https://github.com/openedx/frontend-app-learner-dashboard.git",
         "port": 1996,
     },
     "learning": {
@@ -65,7 +72,9 @@ CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
         "port": 2000,
     },
     "ora-grading": {
-        "repository": "https://github.com/openedx/frontend-app-ora-grading.git",
+        "repository": "https://github.com/regisb/frontend-app-ora-grading.git",
+        "version": "regisb/rsbuild",
+        # "repository": "https://github.com/openedx/frontend-app-ora-grading.git",
         "port": 1993,
     },
     "profile": {
--- a/tutormfe/templates/mfe/build/mfe/Dockerfile
+++ b/tutormfe/templates/mfe/build/mfe/Dockerfile
@@ -9,7 +9,7 @@ RUN apt update \
     # required for gifsicle, mozjpeg, and optipng (on arm)
     autoconf libtool pkg-config zlib1g-dev \
     # required for node-sass (on arm)
-    python g++ \
+    python3 g++ \
     # required for image-webpack-loader (on arm)
     libpng-dev \
     # required for building node-canvas (on arm, for authoring)
--- a/tutormfe/templates/mfe/build/mfe/Dockerfile
+++ b/tutormfe/templates/mfe/build/mfe/Dockerfile
@@ -17,8 +17,14 @@ RUN apt update \
     libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
 
 RUN mkdir -p /openedx/app /openedx/env
+
+# Install shared dependencies in /openedx/node_modules
+WORKDIR /openedx/
+RUN npm install @rsbuild/core @rsbuild/plugin-react @rsbuild/plugin-svgr @rsbuild/plugin-sass
+
+# Install apps in /openedx/app
 WORKDIR /openedx/app
-ENV PATH=/openedx/app/node_modules/.bin:${PATH}
+ENV PATH=/openedx/app/node_modules/.bin:/openedx/node_modules/.bin:${PATH}
 
 {{ patch("mfe-dockerfile-base") }}
 
@@ -77,7 +83,9 @@ CMD ["/bin/bash", "-c", "npm run start --- --config ./webpack.dev-tutor.config.j
 ######## {{ app_name }} (production)
 FROM {{ app_name }}-common AS {{ app_name }}-prod
 ENV NODE_ENV=production
-RUN npm run build
+COPY rsbuild.config.ts /openedx/app/rsbuild.config.ts
+{# Locked cache will prevent concurrent builds of MFE apps, to spare CPU usage #}
+RUN --mount=type=cache,target=/openedx/app/node_modules/.cache,sharing=locked rsbuild build
 {{ patch("mfe-dockerfile-post-npm-build") }}
 {{ patch("mfe-dockerfile-post-npm-build-{}".format(app_name)) }}
 {% endfor %}
--- a/tutormfe/templates/mfe/build/mfe/rsbuild.config.ts
+++ b/tutormfe/templates/mfe/build/mfe/rsbuild.config.ts
@@ -31,6 +31,7 @@ const appName = parsedEnv.APP_ID || path.basename(currentDir).replace('frontend-
 const publicPath = parsedEnv.PUBLIC_PATH || '/' + appName + '/';
 const dev = parsedEnv.NODE_ENV === 'development';
 const prod = !dev;
+const isArm64 = (process.arch === 'arm64');
 parsedEnv.APP_ID = appName;
 parsedEnv.PUBLIC_PATH = publicPath;
 parsedEnv.MFE_CONFIG_API_URL = prod ? '/api/mfe_config/v1' : 'http://local.openedx.io:8000/api/mfe_config/v1';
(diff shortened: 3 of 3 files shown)


regisb:
> Is there anything you found that doesn't work / isn't supported by rsbuild?

No, I haven't. That's because there's nothing magical with our MFEs: they are just frontend code, bundled as regular js/html/css/images etc. 

That being said, I haven't extensively tested the 10 officially maintained MFEs, so it's quite possible I missed some issues.

> If not, we should just replace it everywhere and not just for tutor-mfe.

I'm not sure anymore about that. While I worked on this project, I discovered [some](https://github.com/openedx/frontend-app-discussions/blob/60da5eafc4138d131097013ee957deb37dfcc144/src/components/TinyMCEEditor.jsx#L38) [places](https://github.com/openedx/frontend-app-learning/blob/85e6e9266d15ad5d5922e20ae767383da7f05489/jest.config.js#L15) [where](https://github.com/openedx/frontend-app-learning/blob/85e6e9266d15ad5d5922e20ae767383da7f05489/.eslintrc.js#L18) the webpack abstraction leaked into the MFE codebase, and that really feels like an anti-pattern.

Instead, I now think that we should remove all of the build code from all MFE. There shouldn't be any MFE-specific build step. There should be a simple, documented standard to build any MFE. And we should be able to run something like `mfe-build --dev ./frontend-app-learning`. This "mfe-build" tool could use rsbuild under the hood, or something else, and we should be able to replace that bundler easily in the future, when a better, shinier bundler inevitably comes out.

hinakhadim:
Awesome! I tested this PR on my system, and it significantly reduces the build time—from 2007 seconds (using npm run build / Webpack) to 754 seconds (using rsbuild).

**With Webpack**: `[+] Building 2007.7s (127/127) FINISHED                     docker-container:doublecpu`
image


**With Rsbuild:**  `[+] Building 754.1s (138/138) FINISHED                        docker-container:doublecpu`
image

I am trying to start(tutor local launch) the system and will give more feedback after `tutor local launch`.

bradenmacdonald:
> Instead, I now think that we should remove all of the build code from all MFE. There shouldn't be any MFE-specific build step. There should be a simple, documented standard to build any MFE. And we should be able to run something like mfe-build --dev ./frontend-app-learning. This "mfe-build" tool could use rsbuild under the hood, or something else, and we should be able to replace that bundler easily in the future, when a better, shinier bundler inevitably comes out.

This makes sense to me, and I think it may even have been the original idea behind `frontend-build`, but things have drifted away from that a bit over time.

Danyal-Faheem:
I tested this out on an Arm64 machine (M1 Macbook) and it seems that the rsbuild step breaks on arm64 native images. Here's the traceback:

```
> [account-prod 2/2] RUN --mount=type=cache,target=/openedx/app/node_modules/.cache,sharing=locked rsbuild build:
0.372 
0.373   Rsbuild v1.3.5
0.373 
1.046 info    Rspack persistent cache enabled (experimental)
4.527 Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
4.527 Message:  created a new `Panic` from: out of range integral type conversion attempted
4.527 Location: index.crates.io-1949cf8c6b5b557f/rancor-0.1.0/src/lib.rs:648
4.527 
4.527 Backtrace omitted.
4.527 
4.527 Run with RUST_BACKTRACE=1 environment variable to display it.
4.527 Run with RUST_BACKTRACE=full to include source snippets.
4.527 Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
4.527 Message:  created a new `Panic` from: out of range integral type conversion attempted
4.527 Location: index.crates.io-1949cf8c6b5b557f/rancor-0.1.0/src/lib.rs:648
4.527 
4.527 Backtrace omitted.
4.527 
4.527 Run with RUST_BACKTRACE=1 environment variable to display it.
4.527 Run with RUST_BACKTRACE=full to include source snippets.
4.563 Aborted
------
Dockerfile:507
--------------------
 505 |     COPY rsbuild.config.ts /openedx/app/rsbuild.config.ts
 506 |     
 507 | >>> RUN --mount=type=cache,target=/openedx/app/node_modules/.cache,sharing=locked rsbuild build
 508 |     
 509 |     
--------------------
ERROR: failed to solve: process "/bin/sh -c rsbuild build" did not complete successfully: exit code: 134
```

I then tried it out by building it using the amd64 platform using the command: `tutor images build mfe --no-cache --docker-arg=--platform=linux/amd64` and it worked perfectly. 

The results were impressive nonetheless even with amd64 emulation with the build layers taking 2-3x less time than when we were using webpack.

regisb:
Thanks for testing @Danyal-Faheem! Would you mind trying out debugging with `RUST_BACKTRACE`, as suggested in the error logs?

DawoudSheraz:
I tested this branch on Macbook M2 and got the same error as Danyal mentioned above https://github.com/overhangio/tutor-mfe/pull/252#issuecomment-2796362845. However, even when I specified the platform (amd/64) in build argument, the build process did not complete and gave the same error. npm connection errors were happening as well, which has been more prominent than other build issues for me (even though I am using max2cpu config). I am adding some stats below:

- Indigo enabled
  - Mfe build with —no-cache
    - 614.9s, failed due to storage size issue
  - Mfe build —no-cache —no-registry-cache
    - Failed due to react compatibility issues in Indigo, 1115.7 seconds
- Indigo Disabled
  - Mfe build —no-cache —no-registry-cache
      - Failed with network error, 795.1s
  - Re-tried but no cache
    - Failed with rsbuild error, 403.1 seconds

regisb:
I was able to reproduce the issue by building on a remote arm64 server:

```
$ tutor images build --docker-arg=--platform=arm64 --docker-arg=--target=learner-dashboard-prod mfe
...
0.427 
0.427   Rsbuild v1.3.9
0.427 
1.165 info    Rspack persistent cache enabled (experimental)
9.090 Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
9.090 Message:  created a new `Panic` from: out of range integral type conversion attempted
9.090 Location: index.crates.io-1949cf8c6b5b557f/rancor-0.1.0/src/lib.rs:648
9.090 
9.090 Backtrace omitted.
9.090 
9.090 Run with RUST_BACKTRACE=1 environment variable to display it.
9.090 Run with RUST_BACKTRACE=full to include source snippets.
9.254 Aborted (core dumped)
```

The issue disappears when I disable the cache, which is tagged as "experimental". So what I did was to disable caching on arm64. I also filed the following upstream issue: https://github.com/web-infra-dev/rspack/issues/10118

Thus, Apple users should now be able to test this PR again. :eyes: @Danyal-Faheem @DawoudSheraz

Danyal-Faheem:
@regisb I tested it with the updated commits and It's building on arm64 natively now!

The results are also impressive:

Running with a max-parallelism of 2 (I had to do this otherwise the `npm clean-install` step fails due to network issues), The rsbuild steps are 3-4x faster on average than npm build.

Some stats:

### Build Time Comparison

| Component           | Webpack/NPM (s) | rsbuild (s) |
|--------------------|--------------------|-----------------------------|
| Full MFE Build     | 1228.1             | 748.7                       |
| Learner-Dashboard  | 111.9              | 48.5                        |
| Learning           | 136.3              | 11.1                        |
| Profile            | 86.0               | 13.9                        |

Abdul-Muqadim-Arbisoft:
@arbrandes have we decided a path for this , are we going with vite or rs build? so we can look into this pr accordingly

arbrandes:
We have an epic for this: https://github.com/openedx/frontend-base/issues/130.   However, I believe we should wait for after all MFEs are converted to frontend-base apps and the legacy ones removed, to make our lives (much) easier.

PR #253, opened 2025-04-30
fix: ENABLE_UNIT_PAGE flag not being used during production in MFE_CONFIG

Danyal-Faheem:
After #249 was merged, the setting `ENABLE_UNIT_PAGE` was enabled for the MFE_CONFIG but was only added to the openedx-lms-development-setting patch. This means that this setting is never enabled when running in tutor.

https://github.com/overhangio/tutor-mfe/blob/ae22a8a840b4a01b682fc73c27d8dcd6fded5f81/tutormfe/patches/openedx-lms-development-settings#L44

We now add it to the openedx-lms-production-setting patch as well to make it work for both development and production.


arbrandes (approved):
Good catch.  👍🏼

regisb:
I blame myself for not catching that during review!

PR #254, opened 2025-05-06
chore: add console error logging in env.config.jsx template

DawoudSheraz:
Edly recently faced an issue with Teak sandbox in which env.config.jsx did not apply when both Indigo and Aspects plugins were enabled. Upon debugging, it was found that there was an issue in env.config.jsx related patches in Indigo which caused this. However, it was not apparent at first that this was the issue and it required debugging and re-building images to pinpoint the issue. While we are fixing the Indigo issue in https://github.com/overhangio/tutor-indigo/pull/148, we are adding a small change in tutor-mfe to log the error in env.config.jsx if config fails to apply for any reason. This will make future debugging easier.

### DIFF
--- a/tutormfe/templates/mfe/build/mfe/env.config.jsx
+++ b/tutormfe/templates/mfe/build/mfe/env.config.jsx
@@ -43,7 +43,7 @@ async function setConfig () {
     {%- endfor %}
 
     {{- patch("mfe-env-config-runtime-final") }}
-  } catch (error) { console.error("env.config.jsx failed to apply.";}
+  } catch (err) { console.error("env.config.jsx failed to apply.", err);}
 
   return config;
 }
--- a/tutormfe/templates/mfe/build/mfe/env.config.jsx
+++ b/tutormfe/templates/mfe/build/mfe/env.config.jsx
@@ -43,7 +43,7 @@ async function setConfig () {
     {%- endfor %}
 
     {{- patch("mfe-env-config-runtime-final") }}
-  } catch { }
+  } catch (error) { console.error("env.config.jsx failed to apply.";}
 
   return config;
 }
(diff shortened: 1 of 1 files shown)


tutormfe/templates/mfe/build/mfe/env.config.jsx
```diff
@@ -43,7 +43,7 @@ async function setConfig () {
     {%- endfor %}
 
     {{- patch("mfe-env-config-runtime-final") }}
-  } catch { }
+  } catch (error) { console.error("env.config.jsx failed to apply.";}
```
arbrandes, line 46: Since we're using a catch-all, it would be useful to log the error message as well, I think.

arbrandes (changes requested):
I usually object to `console.log`ging in production code, but it seems plugin configuration and interaction will always be prone to errors.  We can discuss a way to do conditional logging later, but for now I think this is worth it.

I do think we should log the error message as well, though.

tutormfe/templates/mfe/build/mfe/env.config.jsx
DawoudSheraz: Sure. I misunderstood console.error that it would automatically log the error, it is not the case.

tutormfe/templates/mfe/build/mfe/env.config.jsx
```diff
@@ -43,7 +43,7 @@ async function setConfig () {
     {%- endfor %}
 
     {{- patch("mfe-env-config-runtime-final") }}
-  } catch { }
+  } catch (err) { console.error("env.config.jsx failed to apply.", err);}
```
arbrandes, line 46: A small nit:

```suggestion
  } catch (err) { console.error("env.config.jsx failed to apply: ", err);}
```

This will lead the user to expect more information after the hard-coded message.

arbrandes (approved):
Approved, with a small nit.

PR #282, opened 2026-02-18
[WIP] Build process for frontendbase apps

holaontiveros:
## Basic details
This is the first step for the build process of the frontend apps and to allow

- Adds a new filter that will be used to know which apps should be frontendbase
- Adds a patch to prevent translations being pulled for template site (while I figure out how that should work)
- Modify a couple of utility functions to allow frontent-template-site to be mounted (while we agree on specific semantics)
- Modifies Caddyfile so if the app is a frontentbase it directs to template-site instead of the old app folder

## Assumptions
This only uses a single build for template-site, the other MFES with a frontendbase branch are not directly pulled during the build, so you need to be sure your packages.json for the template-site has everything to be able to be build.

If the packages that are intented to be used (the apps to be used in the apps array) are not published (so no dist folder yet) you can point to a git branch and they need to have a package.json [prepare script ](https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-scripts) and [include the right files](https://github.com/WGU-Open-edX/frontend-app-authn/blob/prepare-test/package.json#L16) for it to be build as part of the npm install.

This way you'll endup with the package installed, transpiled and ready to be used, and this magic will only happen when the package is installed from a git repo, which means the build / install dev and normal dependencies will only happen when you need to use a non-published package.

Also for now, the aggregator repo needs to have the same public path as the app_name for the app in this case and for now that's template-site which means `PUBLIC_PATH=/template-site/` because that will allow all the assets to be served properly.

## How to add a frontendapp

```py

@FRONTEND_APPS.add()
def _add_frontend_apps(apps):
    apps["learner-dashboard"] = {}
    apps["authn"] = {}
    apps["instructor"] = {} 
    return apps
```

Any app that it's on this list will be considered a fronend-app so when Caddy shows the content for it, it will show the one on template-site instead of the normal app.

### DIFF
--- a/tutormfe/hooks.py
+++ b/tutormfe/hooks.py
@@ -11,7 +11,13 @@ import typing as t
 from tutor.core.hooks import Filter
 
 MFE_ATTRS_TYPE = t.Dict[t.Literal["repository", "port", "version"], t.Union["str", int]]
+FRONTEND_TEMPLATE_SITE_ATTRS_TYPE = t.Dict[t.Literal["repository", "version"], t.Union["str", int]]
 
 MFE_APPS: Filter[dict[str, MFE_ATTRS_TYPE], []] = Filter()
+
+# This will hold the frontent-template-site like repos (ideally one)
+FRONTEND_BASE_APPS: Filter[dict[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE], []] = Filter()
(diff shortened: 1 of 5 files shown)


arbrandes:
Closing in favor of https://github.com/overhangio/tutor-mfe/pull/284.

PR #283, opened 2026-03-10
[WIP] Build process for frontendbase apps alternative explicit version

holaontiveros:
## Basic details
This is the first step for the build process of the frontend apps

- Adds a new filter that will be used to know which apps should be frontendbase
- Adds a patch to prevent translations being pulled for template site (while I figure out how that should work)
- Modify a couple of utility functions to allow frontent-template-site to be mounted (while we agree on specific semantics)
- Modifies Caddyfile so if the app is a frontentbase it directs to template-site instead of the old app folder
- installs all the frontendapps that are configured with a repository attribute which means they need to be build explicitly, pack and installed on template site

## Assumptions
This use multi stage builds if there's frontend apps that have the repository attribute configured which means that if you manually declare a repository for those apps the process will try to build it, pack it and install it.

** This may be network / RAM intensive ** so in case you run into something like:

image

disable a couple MFEs or reduce the amount of parallelism in docker build config (details about this can be found at the readme [look for parallelism in the dev section](https://github.com/overhangio/tutor-mfe?tab=readme-ov-file#mfe-development))

Also for now, the aggregator repo needs to have the same public path as the app_name for the app in this case and for now that's template-site which means `PUBLIC_PATH=/template-site/` because that will allow all the assets to be served properly.

## How to add a frontendapp

```py

@FRONTEND_APPS.add()
def _add_frontend_apps(apps):
    apps["learner-dashboard"] = {}
    apps["authn"] = {}
    apps["instructor"] = {} 
    return apps
```

and if you need some of those to be manually build and installed in the template site:

```py
@FRONTEND_APPS.add()
def _add_frontend_apps(apps):
    apps["learner-dashboard"] = {
        "repository": "https://github.com/openedx/frontend-app-learner-dashboard.git",
        "version": "frontend-base",
        "port": "8080",
    }
    apps["authn"] = {
        "repository": "https://github.com/openedx/frontend-app-authn.git",
        "version": "frontend-base",
        "port": "8080",
    }
    return apps
```

_notice that the `port`is there but it's just to match the contract for the Dockerfile_ 

Any app that it's on this list will be considered a fronend-app so when Caddy shows the content for it, it will show the one on template-site instead of the normal app.

### DIFF
--- a/tutormfe/hooks.py
+++ b/tutormfe/hooks.py
@@ -11,7 +11,13 @@ import typing as t
 from tutor.core.hooks import Filter
 
 MFE_ATTRS_TYPE = t.Dict[t.Literal["repository", "port", "version"], t.Union["str", int]]
+FRONTEND_TEMPLATE_SITE_ATTRS_TYPE = t.Dict[t.Literal["repository", "version"], t.Union["str", int]]
 
 MFE_APPS: Filter[dict[str, MFE_ATTRS_TYPE], []] = Filter()
+
+# This will hold the frontent-template-site like repos (ideally one)
+FRONTEND_BASE_APPS: Filter[dict[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE], []] = Filter()
+# This will hold the list of which apps are "enabled" so we can switch between mfe and frontend-base ones
+FRONTEND_APPS: Filter[dict[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE], []] = Filter()
 
 PLUGIN_SLOTS: Filter[list[tuple[str, str, str]], []] = Filter()
--- a/tutormfe/hooks.py
+++ b/tutormfe/hooks.py
@@ -11,11 +11,12 @@ import typing as t
 from tutor.core.hooks import Filter
 
 MFE_ATTRS_TYPE = t.Dict[t.Literal["repository", "port", "version"], t.Union["str", int]]
-FRONTEND_TEMPLATE_SITE_ATTRS_TYPE = t.Dict[t.Literal["repository", "version"], t.Union["str", int]]
+FRONTEND_TEMPLATE_SITE_ATTRS_TYPE = t.Dict
 
 MFE_APPS: Filter[dict[str, MFE_ATTRS_TYPE], []] = Filter()
 
-# TODO: This will hold the list of which apps are "enabled" so we can switch between mfe and frontend-base ones
-FRONTEND_APPS: Filter[dict[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE], []] = Filter()
+# TODO: This will hold the list of which apps are "enabled" so we can switch between mfe
+# and frontend-base ones
+FRONTEND_APPS: Filter[dict[str, t.Dict], []] = Filter()
 
 PLUGIN_SLOTS: Filter[list[tuple[str, str, str]], []] = Filter()
--- a/tutormfe/hooks.py
+++ b/tutormfe/hooks.py
@@ -15,9 +15,7 @@ FRONTEND_TEMPLATE_SITE_ATTRS_TYPE = t.Dict[t.Literal["repository", "version"], t
 
 MFE_APPS: Filter[dict[str, MFE_ATTRS_TYPE], []] = Filter()
 
-# This will hold the frontent-template-site like repos (ideally one)
-FRONTEND_BASE_APPS: Filter[dict[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE], []] = Filter()
-# This will hold the list of which apps are "enabled" so we can switch between mfe and frontend-base ones
+# TODO: This will hold the list of which apps are "enabled" so we can switch between mfe and frontend-base ones
 FRONTEND_APPS: Filter[dict[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE], []] = Filter()
 
 PLUGIN_SLOTS: Filter[list[tuple[str, str, str]], []] = Filter()
--- a/tutormfe/hooks.py
+++ b/tutormfe/hooks.py
@@ -10,13 +10,17 @@ import typing as t
 
 from tutor.core.hooks import Filter
 
-MFE_ATTRS_TYPE = t.Dict[t.Literal["repository", "port", "version"], t.Union["str", int]]
-FRONTEND_TEMPLATE_SITE_ATTRS_TYPE = t.Dict
+MFE_ATTRS_TYPE = t.Dict[
+    t.Literal["repository", "port", "version"], t.Union["str", int]
+]
+FRONTEND_TEMPLATE_SITE_ATTRS_TYPE = t.Dict[
+    t.Literal["repository", "version"], t.Union["str", int]
+]
 
 MFE_APPS: Filter[dict[str, MFE_ATTRS_TYPE], []] = Filter()
 
 # TODO: This will hold the list of which apps are "enabled" so we can switch between mfe
 # and frontend-base ones
-FRONTEND_APPS: Filter[dict[str, t.Dict], []] = Filter()
+FRONTEND_APPS: Filter[dict[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE], []] = Filter()
 
 PLUGIN_SLOTS: Filter[list[tuple[str, str, str]], []] = Filter()
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -115,17 +115,16 @@ def get_frontend_apps(apps_to_build: bool = False) -> dict[str, FRONTEND_TEMPLAT
     all_frontend_apps = FRONTEND_APPS.apply({})
 
     if not apps_to_build:
-        return {}
+        return all_frontend_apps
 
     # When returning apps to build we only return the ones that have a repository defined
     # (those are the ones to be built) and we prefix the name
     # with "frontend-app-" to avoid conflicts with MFE names
-    apps_to_return = {
+    return {
         f"frontend-app-{name}": attrs 
         for name, attrs in all_frontend_apps.items() 
         if "repository" in attrs
     }
-    return apps_to_return
 
 
 @tutor_hooks.lru_cache
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -13,7 +13,7 @@ from tutor.hooks import priorities
 from tutor.types import Config, get_typed
 
 from .__about__ import __version__
-from .hooks import MFE_APPS, MFE_ATTRS_TYPE, FRONTEND_BASE_APPS, FRONTEND_APPS, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE, PLUGIN_SLOTS
+from .hooks import MFE_APPS, MFE_ATTRS_TYPE, FRONTEND_APPS, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE, PLUGIN_SLOTS
 
 # Handle version suffix in main mode, just like tutor core
 if __version_suffix__:
@@ -78,8 +78,8 @@ CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
     },
     "template-site": {
         "repository": "https://github.com/WGU-Open-edX/frontend-template-site.git",
+        "version": "initial",
         "port": 8080,
-        "frontend_base_app": True,  # Mark this app as a frontend base app maybe (?)
     } 
 }
 
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -5,7 +5,7 @@ import typing as t
 from glob import glob
 
 import importlib_resources
-from tutor import fmt
+from tutor import fmt, config as tutor_config
 from tutor import hooks as tutor_hooks
 from tutor.__about__ import __version_suffix__
 from tutor.bindmount import iter_mounts
@@ -13,7 +13,7 @@ from tutor.hooks import priorities
 from tutor.types import Config, get_typed
 
 from .__about__ import __version__
-from .hooks import MFE_APPS, MFE_ATTRS_TYPE, PLUGIN_SLOTS
+from .hooks import MFE_APPS, MFE_ATTRS_TYPE, FRONTEND_BASE_APPS, FRONTEND_APPS, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE, PLUGIN_SLOTS
 
 # Handle version suffix in main mode, just like tutor core
 if __version_suffix__:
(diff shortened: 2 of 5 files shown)


tutormfe/templates/mfe/apps/mfe/Caddyfile
```diff
@@ -26,15 +26,22 @@
...
-    {% for app_name, app in iter_mfes() %}
+    {% for app_name, app in iter_unique_apps() %}
     @mfe_{{ app_name }} {
         path /{{ app_name }} /{{ app_name }}/*
     }
     handle @mfe_{{ app_name }} {
         uri strip_prefix /{{ app_name }}
+        {%- if is_frontend_app_enabled(app_name) %}
+        # {{ app_name }} - using frontend-apps approach
+        root * /openedx/dist/template-site
```
arbrandes, line 37: How about calling it `frontend-site`?

```suggestion
        root * /openedx/dist/frontend-site
```

tutormfe/hooks.py
```diff
@@ -10,8 +10,17 @@
 
 from tutor.core.hooks import Filter
 
-MFE_ATTRS_TYPE = t.Dict[t.Literal["repository", "port", "version"], t.Union["str", int]]
+MFE_ATTRS_TYPE = t.Dict[
+    t.Literal["repository", "port", "version"], t.Union["str", int]
+]
+FRONTEND_TEMPLATE_SITE_ATTRS_TYPE = t.Dict[
```
arbrandes, line 16: Unless I misunderstood the purpose, this is about individual apps, not the whole site, right?

```suggestion
FRONTEND_APP_ATTRS_TYPE = t.Dict[
```

tutormfe/plugin.py
```diff
@@ -76,6 +76,11 @@
         "repository": "https://github.com/openedx/frontend-app-profile.git",
         "port": 1995,
     },
+    # "template-site": {
+    #     "repository": "https://github.com/WGU-Open-edX/frontend-template-site.git",
+    #     "version": "initial",
+    #     "port": 8080,
+    # }
```
arbrandes, line 83: Can we remove this commented out bit?

tutormfe/plugin.py
```diff
@@ -95,6 +100,49 @@ def get_mfes() -> dict[str, MFE_ATTRS_TYPE]:
     return MFE_APPS.apply({})
 
 
+# List will need
+## Apps that are only frontend-apps
+## Apps that are only MFEs
+## Apps with unique ones (all old mfes + instruct)
+## 1 and 2 with 1 having something like a different identifier
```
arbrandes, line 107: Without reading the code, this comment is hard to understand.  What are "apps with unique ones"?  What are 1 and 2?

tutormfe/plugin.py
```diff
@@ -95,6 +100,49 @@ def get_mfes() -> dict[str, MFE_ATTRS_TYPE]:
...
+
+
+@tutor_hooks.lru_cache
+def get_frontend_apps(apps_to_build: bool = False) -> dict[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE]:
```
arbrandes, line 111: `apps_to_build` sounds like a list, but apparently it's a boolean.  Let's call it something like `only_apps_to_build`, then?

```suggestion
def get_frontend_apps(only_apps_to_build: bool = False) -> dict[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE]:
```

tutormfe/plugin.py
```diff
@@ -95,6 +100,49 @@ def get_mfes() -> dict[str, MFE_ATTRS_TYPE]:
...
+
+@tutor_hooks.lru_cache
+def get_all_apps() -> dict[str, t.Union[MFE_ATTRS_TYPE, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE]]:
+    """
+    This function is cached for performance.
+    """
+    # IMPORTANT: Make a copy to avoid mutating the cached result from get_frontend_apps()
+    all_apps = get_frontend_apps(apps_to_build=True).copy()
+    mfes = get_mfes()
+    all_apps.update(mfes)
```
arbrandes, line 138: Nit: this makes it feel like the frontend apps are "more default", when it's really the other way around.  For now, I think we should populate the list with the regular MFEs first.  Also, I feel like we should distinguish more clearly between MFEs and apps.


```suggestion
    mfes_and_apps = get_mfes()
    apps_to_build = get_frontend_apps(only_apps_to_build=True).copy()
    mfes_and_apps.update(apps_to_build)
```

tutormfe/plugin.py
```diff
@@ -95,6 +100,49 @@ def get_mfes() -> dict[str, MFE_ATTRS_TYPE]:
...
+
+    # ensure frontend-template-site is the last one on the all_apps dict
+    if "template-site" in all_apps:
+        all_apps["template-site"] = all_apps.pop("template-site")
```
arbrandes, line 142:
```suggestion
    if "frontend-site" in mfes_and_apps:
        mfes_and_apps["frontend-site"] = mfes_and_apps.pop("frontend-site")
```

tutormfe/plugin.py
```diff
@@ -95,6 +100,49 @@ def get_mfes() -> dict[str, MFE_ATTRS_TYPE]:
...
+    # with "frontend-app-" to avoid conflicts with MFE names
+    return {
+        f"frontend-app-{name}": attrs 
+        for name, attrs in all_frontend_apps.items() 
+        if "repository" in attrs
+    }
+
+
+@tutor_hooks.lru_cache
+def get_all_apps() -> dict[str, t.Union[MFE_ATTRS_TYPE, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE]]:
```
arbrandes, line 131: This is not literally getting all apps.  Just the ones that need to be build locally:

```suggestion
def get_mfes_and_apps_to_build() -> dict[str, t.Union[MFE_ATTRS_TYPE, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE]]:
```

tutormfe/plugin.py
```diff
@@ -129,6 +177,50 @@ def iter_mfes() -> t.Iterable[tuple[str, MFE_ATTRS_TYPE]]:
     """
     yield from get_mfes().items()
 
+# Iter throgh all mfes and adds the unique frontend apps,
+# so we can have a list of all the things that are unique that needs
+# to be added to Caddyfile, for example instructor dashboard that was 
+# created as frontend-base app but didn't exist as a MFE before
+# so it returns the whole mfes list plus the unique frontend apps that are not in the mfe list
```
arbrandes, line 184: I had to read the code to understand why the choice of "unique", here.  I think it would be a good idea to explicitly describe the situation where there might be duplication, including an example.

Also, since is just for the Caddyfile, let's call the function what it is: `iter_paths`.

tutormfe/plugin.py
```diff
@@ -129,6 +177,50 @@ def iter_mfes() -> t.Iterable[tuple[str, MFE_ATTRS_TYPE]]:
...
+def iter_unique_apps() -> t.Iterable[tuple[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE]]:
```
arbrandes, line 185:
```suggestion
def iter_paths() -> t.Iterable[tuple[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE]]:
```

tutormfe/templates/mfe/apps/mfe/Caddyfile
```diff
@@ -26,15 +26,22 @@
     redir @authoring /authoring/{re.authoring.1} permanent
     {% endif %}
 
-    {% for app_name, app in iter_mfes() %}
+    {% for app_name, app in iter_unique_apps() %}
```
arbrandes, line 29: (See the other comment about the name of this function)

```suggestion
    {% for path, app in iter_paths() %}
```

tutormfe/templates/mfe/apps/mfe/Caddyfile
```diff
@@ -26,15 +26,22 @@
...
     @mfe_{{ app_name }} {
         path /{{ app_name }} /{{ app_name }}/*
     }
     handle @mfe_{{ app_name }} {
         uri strip_prefix /{{ app_name }}
+        {%- if is_frontend_app_enabled(app_name) %}
```
arbrandes, line 35: Shorter and clearer, I think:

```suggestion
        {%- if is_frontend_app(path) %}
```

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -22,7 +22,7 @@ ENV PATH=/openedx/app/node_modules/.bin:${PATH}
 
 {{ patch("mfe-dockerfile-base") }}
 
-{% for app_name, app in iter_mfes() %}
+{% for app_name, app in iter_all_apps() %}
```
arbrandes, line 25: For clarity, as suggested elsewhere:

```suggestion
{% for app_name, app in iter_mfes_and_apps_to_build() %}
```

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -50,9 +50,11 @@ RUN --mount=type=cache,target=/root/.npm,sharing=shared npm clean-install --no-a
 {{ patch("mfe-dockerfile-post-npm-install-{}".format(app_name)) }}
 COPY --from={{ app_name }}-src / /openedx/app
 
+# While we figure out how translations will be managed  in template site
+# if it's either template-site or starts with frontend-app, we skip pulling translations
+{% if app_name != "template-site" and not app_name.startswith("frontend-app") %}
```
arbrandes, line 55: I'm not a fan of distinguishing between MFE and frontend app just based on a string prefix.  I think we should own up to the difference and have a separate loop just for frontend apps.  For example, I suspect apps don't need the pre- and post-npm-install patches.

Also, I don't think frontend apps will be subject to the same MFE_COMMON_VERSION (we're not tagging frontend-base apps with `release/verawood.1`).  This is another reason to have a separate list just for them.

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -66,6 +68,12 @@ COPY env.config.jsx /openedx/app
 {{ patch("mfe-dockerfile-pre-npm-build") }}
 {{ patch("mfe-dockerfile-pre-npm-build-{}".format(app_name)) }}
 
+{% if is_frontend_app_to_build(app_name) %}
```
arbrandes, line 71: `is_frontend_app()` should be sufficient because this list only contains things to build already, but as noted in the above comment, I actually think we should have a separate loop just for frontend apps and avoid the check altogether.

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -76,6 +84,16 @@ CMD ["/bin/bash", "-c", "npm run start --- --config ./webpack.dev-tutor.config.j
 {%- for app_name, app in iter_mfes() %}
 ######## {{ app_name }} (production)
 FROM {{ app_name }}-common AS {{ app_name }}-prod
+
+{% if app_name == "template-site" %}
```
arbrandes, line 88: The frontend site should not be an item on a configurable list.  It should be a hard-coded part of the Dockerfile.

arbrandes (changes requested):
Thanks, Javier!

In this first pass, I made several inline comments, but as I write this some of them are already obsolete lol.   Anyway, the gist of it comes down to:

1. We should not depend on an arbitrary prefix to distinguish frontend apps from regular MFEs.  They should be "their own thing" as much as possible.
2. We need a separate section/loop of the Dockerfile from the one that builds regular MFEs, because the requirements and expectations, though similar, are not quite the same.
3. tutor-mfe needs its own minimal version of frontend-template-site.  As in, it should be _in this repository_, not in a separate one.  We'd just call it `frontend-site`, or something like that.  This is because we're going to have to let Tutor customize certain files via patches: notably, `package.json` and `index.html`.

And now the important bit:

4. NPM workspaces: I believe we can reconcile the explicit build being proposed here (which I prefer) with some of the advantages of the `npm prepare` approach you suggest in #282 with NPM workspaces - but even better.  The basic plan is the following:

### package.json differences from frontend-template-site

Add a `workspaces` field and add the build:deps script:

```json
{
  "workspaces": [
    "packages/*"
  ],
  "scripts": {
    "build:deps": "npm run build --workspaces --if-present",
  }
}
```

Dependencies like `@openedx/frontend-app-learner-dashboard` remain listed under `dependencies` with their registry versions. When a matching package exists under `packages/`, npm resolves the dependency via the workspace checkout instead of fetching from the registry.

### Local checkouts

Apps that need building (and only apps that need building) are checked out into `frontend-site/packages/`.  This can be done in the aforementioned separate loop in the Dockerfile, all in a single layer.

When no packages are checked out, `npm install` falls back to the registry as usual — the same `package.json` works both ways.

### Build Workflow

This is the single install/build step in the Dockerfile after everything is checked out:

```bash
npm install                     # hoists and deduplicates all dependencies, including devDependencies like tsc and tsc-alias
npm run build:deps                   # builds the workspace packages (and just those)
npm run build            # build the site
```

`npm run build --workspaces` runs each package's `build` script in topological order. `--if-present` skips packages that don't have one.

### Advantages Over `prepare` in Each Package

- **Shared build tools**: `tsc`, `tsc-alias`, and other build dependencies are installed once at the root, not independently in each package. Faster installs, less disk usage.
- **Explicit build step**: `build:deps` is a deliberate action, not a side effect of `npm install`. Easier to reason about and debug.
- **No devDependency churn**: With `prepare` on git dependencies, npm installs each package's devDependencies, runs the build, then prunes them. Workspaces skip this install-build-prune cycle entirely.
- **Graceful fallback**: Without `packages/` checkouts, the project installs normally from the registry.

tutormfe/templates/mfe/apps/mfe/Caddyfile
holaontiveros: Agree, this is currently based on the naming for the especific repo, so given that it's called "template-site" it ends up with that name, so it's a matter of changing the config

tutormfe/templates/mfe/build/mfe/Dockerfile
holaontiveros: We can definitely separate them, I just didn't want to start with something separated without having everything rolling and being able to know how we can "optimize it"

I mean so the new separate loop does only the things we know we want it to do

tutormfe/templates/mfe/build/mfe/Dockerfile
holaontiveros: So we'll have package.json, and a site.config.build.tsx tempaltes that we'll use as a base always and patch on top of it?

tutormfe/plugin.py
holaontiveros: Agree, It was for my own read when I was iteraring through different approachces, I'll clean it as soon as we get to the final draft

tutormfe/plugin.py
holaontiveros: I had it to remember to add the template site as a default app, BUT if we are going with base files to be patched then yeah I'll just kill this part

arbrandes:
Forgot an important bit:

I don't know if we'll have to build the list of dependencies in the frontend-site package.json dynamically from the plugin data, or if we can manage with separate `npm install`s.  My hunch is that building the package.json dynamically (from package names and loose versions such as `^1`) will be cleaner.   We can still give users `pre-` and `post-npm-install` patches, though.

arbrandes:
Closing in favor of https://github.com/overhangio/tutor-mfe/pull/284.

PR #284, opened 2026-03-26
[WIP] Build process for frontendbase apps npm workspaces flavor (aka: the good one)

holaontiveros:
## Basic details
This is the first step for the build process of the frontend apps, what it does:

- Adds a new filter (`FRONTEND_APPS`) that will be used to know which apps should be frontendbase
- Adds a new filter (`FRONTEND_SITES`) to know which "sites" are going to be build, by default we have a local one but operators are going to be able to override the default one or add extra ones
- Adds CORS patches for the configured sites
- Adds general functions to iterate and get the right information for the new filters
- Adds a utility `iter_paths` which returns the paths that should be used in the Caddyfile, taking in account that they could be either from the old MFEs or from a new frontendapp that didn't have MFE before
- Adds utility `is_frontend_app` which checks if the app is "enabled" as a frontend app so Caddy does some path magic
- Caddyfile know checks if the path is for a frontend enabled app so it returns the right folder for it based on config and serves all possible frontend-sites
- Adds all the basic files for a frontend-site to be build
  - The files have templating so any configured frontendapp is added dynamically to `package.json` and `site.scss`
  - templating for right domains in `site.config.x.tsx`

## Pending things
- Figure the mounting things for sites and frontend apps
- use right site.config based on being dev or prod (?)
- ~~Maybe multi-stage sites part (?)~~
- ~~General cleanup and better semantic namings~~ 
- ~~entryPoint configs for apps~~ 
- ~~configs for site~~ 
- ~~pluginSlots equivalent~~
- ~~patch locations~~
- final readme with the details after we have everything else

## How to add a frontendapp

```py

@FRONTEND_APPS.add()
def _add_frontend_apps(apps):
    apps["learner-dashboard"] = {
    "appEntryPoints": {
            "packageName": "@openedx/frontend-app-learner-dashboard",
            "packageVersion": "^1.0.0-alpha",
            "styles": ["app.scss"],
            "components": ["learnerDashboardApp"],
        },
    }
    apps["authn"] = {
    "appEntryPoints": {
            "packageName": "@openedx/frontend-app-authn",
            "packageVersion": "^1.0.0-alpha",
            "styles": ["app.scss"],
            "components": ["authnApp"],
        },
    }
    apps["instructor"] = {
    "appEntryPoints": {
            "packageName": "@openedx/frontend-app-instruct",
            "packageVersion": "^1.0.0-alpha",
            "styles": ["app.scss"],
            "components": ["instructApp"],
        },
    } 
    return apps
```

and if you need some of those to be manually build and installed in the frontend-site:

```py
@FRONTEND_APPS.add()
def _add_frontend_apps(apps):
    apps["learner-dashboard"] = {
        "repository": "https://github.com/arbrandes/frontend-app-learner-dashboard.git",
        "version": "workspaces-dev",
        "appEntryPoints": {
            "packageName": "@openedx/frontend-app-learner-dashboard",
            "packageVersion": "^1.0.0-alpha",
            "styles": ["app.scss"],
            "components": ["learnerDashboardApp"],
        },
    }
    apps["authn"] = {
        "repository": "https://github.com/arbrandes/frontend-app-authn.git",
        "version": "workspaces-dev",
        "appEntryPoints": {
            "packageName": "@openedx/frontend-app-authn",
            "packageVersion": "^1.0.0-alpha",
            "styles": ["app.scss"],
            "components": ["authnApp"],
        },
    }

    return apps
```

When no `repository`and `version`are specified nothing will be added to the `packages/` folder for that frontend-app so npm will try to fetch them from the registry

Any app that it's on this list will be considered a fronend-app so when Caddy shows the content for it, it will show the one on frontend-site instead of the normal app.

By default there's going to be a frontend-site that it's build from files in the tutor-mfe plugin (everything under `tutormfe/templates/mfe/build/mfe/frontend-site`)

if the operator wants to add a new site or substitute the default one it can be done like:

```py
@FRONTEND_SITES.add()
def _add_frontend_base_apps(sites):
    sites['default'] = {
      "repository": "https://github.com/openedx/frontend-site-custom",
      "version": "main"
      "port": "8080"
    }
    return sites
```

The structure is the same as adding an MFE the only thing to take in account is that if you want to add a new site with a different name than default you also need to specify which frontend apps will live there so Caddy knows that it should serve those files

```py
@FRONTEND_SITES.add()
def _add_frontend_base_apps(sites):
    sites['new-site'] = {
      "repository": "https://github.com/openedx/frontend-site-custom",
      "version": "main"
      "port": "8082"
    }
    return sites

@FRONTEND_APPS.add()
def _add_frontend_apps(apps):
    apps["learner-dashboard"] = {
        "repository": "https://github.com/openedx/frontend-app-learner-dashboard.git",
        "version": "frontend-base",
        "site": "new-site"
    }
    return apps
```

In case that further customization is required for the a frontensite the full config looks like:

```py
@FRONTEND_SITES.add()
def _add_frontend_base_apps(sites):
    sites['default'] = {
        "repository": "local",
        "port": 8080,
        "siteConfig": {
            "siteId": "tutor-frontend-site",
            "siteName": "Frontend Template Site",
            "accessTokenCookieName": "edx-jwt-cookie-header-payload",
            "redirectRoleId": "org.openedx.frontend.role.dashboard",
            "frontendBaseVersion": "^1.0.0-alpha",
            "paragonVersion": "^23", 
            "externalRoutes": [
                {
                    "role": "org.openedx.frontend.role.profile",
                    "url": ":1995/profile/",
                },
                {
                    "role": "org.openedx.frontend.role.account",
                    "url": ":1997/account/",
                },
                {
                    "role": "org.openedx.frontend.role.logout",
                    "url": ":8000/logout",
                },
            ],
        },
    },

    return sites
```

### Config values reference 

| value | description | default |
|--------|--------|--------|
| siteId | direct equivalent of `siteId` inside site.config.xx.tsx | `'tutor-frontend-site'`|
| siteName | direct equivalent of `siteName` inside site.config.xx.tsx | `'Frontend site'`|
| accessTokenCookieName | direct equivalent of `accessTokenCookieName` inside site.config.xx.tsx | `'edx-jwt-cookie-header-payload'`|
| redirectRoleId | role id to be used to readirect navigation to the default root route when someone access `http://apps.local.openedx.io:8080/` (root route) | `''` the default would fail and throw a 404 so a role needs to always be defined and pointing to the proper place |
| frontendBaseVersion | allows the configuration of a custom version of frontend-base or assign an [alias](https://docs.npmjs.com/cli/v8/using-npm/package-spec#aliases) for it | `'tutor-frontend-site'`|
| paragonVersion | allows the configuration of a custom version of paragon or assign an [alias](https://docs.npmjs.com/cli/v8/using-npm/package-spec#aliases) for it | `'tutor-frontend-site'`|
| externalRoutes | allows the configuration of the external routes to be configured at site.config.xx.tsx | `[]`|

## Front end slots
Given the current architecture of a frontendbase the slots part has been dramatically simplified and made more powerful, so for complex slot situations you can always create an app based on the [template-site](https://github.com/openedx/frontend-template-site) and just export the slots you require, but if you need a single file plugin / config for a slot then we get the new `@FRONTEND_SLOTS`

The usage will be similar to the current `@PLUGIN_SLOTS` but way more simplified, ex:

```py
@FRONTEND_SLOTS.add()
def _add_app_slots(app_slots):
    app_slots.append(
        (
            "org.openedx.frontend.slot.header.primaryLinks.v1",
            """
            {
                slotId: 'org.openedx.frontend.slot.header.primaryLinks.v1',
                id: 'org.openedx.frontend.widget.learnerDashboard.headerLinkDiscover.v999',
                op: WidgetOperationTypes.APPEND,
                element: (
                    <a href="https://github.com/openedx/frontend-base/blob/main/docs/decisions/0009-slot-naming-and-lifecycle.rst"
                    className="nav-link"
                    >
                    Discover the power of slots!</a>
                )
            }
            """,
        )
    )
    return app_slots
```

This is a widget that it's configured to add be added at the `org.openedx.frontend.slot.header.primaryLinks.v1'` ID which exists in the header and Adds a new link next to whichever is already there

image

The main difference is that  we no longer need the app name, just the slotID (and currently it's not used...) and the slot config, any slot configured here will be configured in a mini-app which just makes the slots available for the build  `slots.tsx`

## New patches
Some new patches were introduced to cover previous available functionality that may be used to extend frontend app functionality

| name | location | description |
|--------|--------|--------|
| frontend-site-dockerfile-pre-npm-install | Dockerfile | Runs inmediatly before site npm install |
| frontend-site-dockerfile-post-npm-install | Dockerfile | Runs inmediatly after site npm install |
| frontend-site-dockerfile-pre-npm-build | Dockerfile | Runs inmediatly before site npm run build |
| frontend-site-dockerfile-post-npm-build | Dockerfile | Runs inmediatly after site npm run build |
| frontend-slots-env-config-buildtime-imports | slotsApp.tsx | Lives before the definition fo the slotsApp |

### DIFF
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -20,6 +20,7 @@ from .hooks import (
     FRONTEND_APPS,
     FRONTEND_APP_ATTRS_TYPE,
     PLUGIN_SLOTS,
+    FRONTEND_SLOTS,
     SITE_ATTRS_TYPE,
 )
 
@@ -182,6 +183,14 @@ def get_plugin_slots(mfe_name: str) -> list[tuple[str, str]]:
     return [i[-2:] for i in PLUGIN_SLOTS.iterate() if i[0] == mfe_name]
 
 
+@tutor_hooks.lru_cache
+def get_frontend_slots() -> list[tuple[str, str]]:
+    """
+    This function is cached for performance.
+    """
+    return FRONTEND_SLOTS.apply([])
+
+
 def iter_mfes() -> t.Iterable[tuple[str, MFE_ATTRS_TYPE]]:
     """
     Yield:
@@ -241,6 +250,15 @@ def iter_plugin_slots(mfe_name: str) -> t.Iterable[tuple[str, str]]:
     yield from get_plugin_slots(mfe_name)
 
 
+def iter_frontend_slots() -> t.Iterable[tuple[str, str]]:
+    """
+    Yield:
+
+        (slot_name, plugin_config)
+    """
+    yield from get_frontend_slots()
+
+
 def is_mfe_enabled(mfe_name: str) -> bool:
     return mfe_name in get_mfes()
 
@@ -266,6 +284,7 @@ tutor_hooks.Filters.ENV_TEMPLATE_VARIABLES.add_items(
         ("iter_frontend_sites", iter_frontend_sites),
         ("get_frontend_sites", get_frontend_sites),
         ("iter_plugin_slots", iter_plugin_slots),
+        ("iter_frontend_slots", iter_frontend_slots),
         ("is_mfe_enabled", is_mfe_enabled),
         ("is_frontend_app", is_frontend_app),
         ("MFEMountData", MFEMountData),
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -115,17 +115,16 @@ def get_frontend_apps(apps_to_build: bool = False) -> dict[str, FRONTEND_TEMPLAT
     all_frontend_apps = FRONTEND_APPS.apply({})
 
     if not apps_to_build:
-        return {}
+        return all_frontend_apps
 
     # When returning apps to build we only return the ones that have a repository defined
     # (those are the ones to be built) and we prefix the name
     # with "frontend-app-" to avoid conflicts with MFE names
-    apps_to_return = {
+    return {
         f"frontend-app-{name}": attrs 
         for name, attrs in all_frontend_apps.items() 
         if "repository" in attrs
     }
-    return apps_to_return
 
 
 @tutor_hooks.lru_cache
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -13,7 +13,7 @@ from tutor.hooks import priorities
 from tutor.types import Config, get_typed
 
 from .__about__ import __version__
-from .hooks import MFE_APPS, MFE_ATTRS_TYPE, FRONTEND_BASE_APPS, FRONTEND_APPS, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE, PLUGIN_SLOTS
+from .hooks import MFE_APPS, MFE_ATTRS_TYPE, FRONTEND_APPS, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE, PLUGIN_SLOTS
 
 # Handle version suffix in main mode, just like tutor core
 if __version_suffix__:
@@ -78,8 +78,8 @@ CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
     },
     "template-site": {
         "repository": "https://github.com/WGU-Open-edX/frontend-template-site.git",
+        "version": "initial",
         "port": 8080,
-        "frontend_base_app": True,  # Mark this app as a frontend base app maybe (?)
     } 
 }
 
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -95,6 +95,7 @@ CORE_FRONTEND_SITES: dict[str, SITE_ATTRS_TYPE] = {
             "siteName": "Frontend Template Site",
             "accessTokenCookieName": "edx-jwt-cookie-header-payload",
             "redirectRoleId": "org.openedx.frontend.role.dashboard",
+            "frontendBaseVersion": "^1.0.0-alpha",
             "externalRoutes": [
                 {
                     "role": "org.openedx.frontend.role.profile",
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -5,7 +5,7 @@ import typing as t
 from glob import glob
 
 import importlib_resources
-from tutor import fmt, config as tutor_config
+from tutor import fmt
 from tutor import hooks as tutor_hooks
 from tutor.__about__ import __version_suffix__
 from tutor.bindmount import iter_mounts
@@ -20,6 +20,7 @@ from .hooks import (
     FRONTEND_APPS,
     FRONTEND_APP_ATTRS_TYPE,
     PLUGIN_SLOTS,
+    SITE_ATTRS_TYPE,
 )
 
 # Handle version suffix in main mode, just like tutor core
@@ -85,7 +86,7 @@ CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
     },
 }
 
-CORE_FRONTEND_SITES: dict[str, MFE_ATTRS_TYPE] = {
+CORE_FRONTEND_SITES: dict[str, SITE_ATTRS_TYPE] = {
     "default": {
         "repository": "local",
         "port": 8080,
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -5,7 +5,7 @@ import typing as t
 from glob import glob
 
 import importlib_resources
-from tutor import fmt
+from tutor import fmt, config as tutor_config
 from tutor import hooks as tutor_hooks
 from tutor.__about__ import __version_suffix__
 from tutor.bindmount import iter_mounts
@@ -13,7 +13,7 @@ from tutor.hooks import priorities
 from tutor.types import Config, get_typed
 
 from .__about__ import __version__
-from .hooks import MFE_APPS, MFE_ATTRS_TYPE, PLUGIN_SLOTS
+from .hooks import MFE_APPS, MFE_ATTRS_TYPE, FRONTEND_BASE_APPS, FRONTEND_APPS, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE, PLUGIN_SLOTS
 
 # Handle version suffix in main mode, just like tutor core
 if __version_suffix__:
@@ -95,6 +95,14 @@ def get_mfes() -> dict[str, MFE_ATTRS_TYPE]:
     return MFE_APPS.apply({})
 
 
+@tutor_hooks.lru_cache
+def get_frontend_apps() -> dict[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE]:
+    """
+    This function is cached for performance.
+    """
+    return FRONTEND_APPS.apply({})
+
+
 class MFEMountData:
     """Stores categorized mounted and unmounted MFEs."""
 
@@ -129,6 +137,14 @@ def iter_mfes() -> t.Iterable[tuple[str, MFE_ATTRS_TYPE]]:
     """
     yield from get_mfes().items()
 
+def iter_frontent_apps() -> t.Iterable[tuple[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE]]:
+    """
+    Yield:
+
+        (name, dict)
+    """
+    yield from get_frontend_apps().items()
+
 
 def iter_plugin_slots(mfe_name: str) -> t.Iterable[tuple[str, str]]:
     """
@@ -142,6 +158,9 @@ def iter_plugin_slots(mfe_name: str) -> t.Iterable[tuple[str, str]]:
 def is_mfe_enabled(mfe_name: str) -> bool:
     return mfe_name in get_mfes()
 
+def is_frontend_app_enabled(app_name: str) -> bool:
+    return app_name in get_frontend_apps()
+
 
 def get_mfe(mfe_name: str) -> t.Union[MFE_ATTRS_TYPE, t.Any]:
     return get_mfes().get(mfe_name, {})
@@ -152,8 +171,10 @@ tutor_hooks.Filters.ENV_TEMPLATE_VARIABLES.add_items(
     [
         ("get_mfe", get_mfe),
         ("iter_mfes", iter_mfes),
+        ("iter_frontend_apps", iter_frontent_apps),
         ("iter_plugin_slots", iter_plugin_slots),
         ("is_mfe_enabled", is_mfe_enabled),
+        ("is_frontend_app_enabled", is_frontend_app_enabled),
         ("MFEMountData", MFEMountData),
     ]
 )
@@ -217,6 +238,7 @@ with open(
     tutor_hooks.Filters.CLI_DO_INIT_TASKS.add_item(("lms", task_file.read()))
 
 REPO_PREFIX = "frontend-app-"
+FRONTEND_TEMPLATE_SITE_REPO = "frontend-template-site"
 
 
 @tutor_hooks.Filters.COMPOSE_MOUNTS.add()
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -142,13 +142,28 @@ def iter_mfes() -> t.Iterable[tuple[str, MFE_ATTRS_TYPE]]:
     """
     yield from get_mfes().items()
 
-def iter_frontent_apps() -> t.Iterable[tuple[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE]]:
+# Iter throgh all mfes and adds the unique frontend apps,
+# so we can have a list of all the things that are unique that needs
+# to be added to Caddyfile, for example instructor dashboard that was 
+# created as frontend-base app but didn't exist as a MFE before
+# so it returns the whole mfes list plus the unique frontend apps that are not in the mfe list
+def iter_frontend_apps() -> t.Iterable[tuple[str, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE]]:
     """
     Yield:
 
         (name, dict)
     """
-    yield from get_frontend_apps().items()
+    mfes = get_mfes()
+    frontend_apps = get_frontend_apps()
+    
+    # First yield all MFEs
+    for name, attrs in mfes.items():
+        yield (name, attrs)
+    
+    # Then yield frontend apps that are not already MFEs
+    for name, attrs in frontend_apps.items():
+        if name not in mfes:
+            yield (name, attrs)
 
 
 def iter_plugin_slots(mfe_name: str) -> t.Iterable[tuple[str, str]]:
@@ -176,7 +191,7 @@ tutor_hooks.Filters.ENV_TEMPLATE_VARIABLES.add_items(
     [
         ("get_mfe", get_mfe),
         ("iter_mfes", iter_mfes),
-        ("iter_frontend_apps", iter_frontent_apps),
+        ("iter_frontend_apps", iter_frontend_apps),
         ("iter_plugin_slots", iter_plugin_slots),
         ("is_mfe_enabled", is_mfe_enabled),
         ("is_frontend_app_enabled", is_frontend_app_enabled),
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -13,7 +13,14 @@ from tutor.hooks import priorities
 from tutor.types import Config, get_typed
 
 from .__about__ import __version__
-from .hooks import MFE_APPS, MFE_ATTRS_TYPE, FRONTEND_APPS, FRONTEND_TEMPLATE_SITE_ATTRS_TYPE, PLUGIN_SLOTS
+from .hooks import (
+    FRONTEND_SITES,
+    MFE_APPS,
+    MFE_ATTRS_TYPE,
+    FRONTEND_APPS,
+    FRONTEND_APP_ATTRS_TYPE,
+    PLUGIN_SLOTS,
+)
 
 # Handle version suffix in main mode, just like tutor core
 if __version_suffix__:
@@ -76,11 +83,13 @@ CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
         "repository": "https://github.com/openedx/frontend-app-profile.git",
         "port": 1995,
     },
-    # "template-site": {
-    #     "repository": "https://github.com/WGU-Open-edX/frontend-template-site.git",
-    #     "version": "initial",
-    #     "port": 8080,
-    # } 
+}
+
+CORE_FRONTEND_SITES: dict[str, MFE_ATTRS_TYPE] = {
+    "default": {
+        "repository": "local",
+        "port": 8080,
+    },
 }
 
 
@@ -90,6 +99,14 @@ CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
 def _add_core_mfe_apps(apps: dict[str, MFE_ATTRS_TYPE]) -> dict[str, MFE_ATTRS_TYPE]:
     apps.update(CORE_MFE_APPS)
     return apps
+
+
+@FRONTEND_SITES.add(priority=tutor_hooks.priorities.HIGH)
+def _add_core_frontend_sites(
+    sites: dict[str, MFE_ATTRS_TYPE],
+) -> dict[str, MFE_ATTRS_TYPE]:
+    sites.update(CORE_FRONTEND_SITES)
+    return sites
 
 
 @tutor_hooks.lru_cache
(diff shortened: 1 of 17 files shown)


tutormfe/templates/mfe/apps/mfe/Caddyfile
```diff
@@ -26,15 +26,34 @@
...
+        {%- else %}
+        # {{ app_name }} - using traditional MFE approach
         root * /openedx/dist/{{ app_name }}
+        {%- endif %}
         try_files /{path} /index.html
         file_server
     }
     {% endfor %}
+
+    {% for app_name, app in iter_frontend_sites() %}
```
arbrandes:
```suggestion
    {% for site_name, site in iter_frontend_sites() %}
```

And I also think you should move the site loop closer to the mfe loop, since we're separating apps from mfes.  The last loop should be the frontend-apps, after frontend-sites.

tutormfe/templates/mfe/build/mfe/frontend-site/public/index.html
```diff
@@ -0,0 +1,11 @@
+<!doctype html>
+<html lang="en-us">
+  <head>
+      <title>Frontend Template Site</title>
```
arbrandes: This needs to be configurable by the operator.  Via SITE_NAME, if I remember correctly.

tutormfe/templates/mfe/build/mfe/frontend-site/src/site.scss
```diff
@@ -0,0 +1,6 @@
+@use '@openedx/frontend-base/shell/app.scss' as shell;
+{%- for app_name, app_attrs in iter_frontend_apps() %}
+{%- set camel_case_name = app_name.split('-') | map('title') | list %}
+{%- set camel_case_name = camel_case_name[0].lower() + (camel_case_name[1:] | join('')) %}
```
arbrandes: Doing this programatically might be too optimistic.  Plus, this is kinda hard to follow.  I think I'd prefer it if we asked the user to configure the full appId (`org.openedx.frontend.app.learnerDashboard`) explicity per app.  Then we can use the last field here.

tutormfe/templates/mfe/build/mfe/frontend-site/package.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "@openedx/frontend-template-site",
+  "version": "1.0.0",
+  "description": "Frontend site template",
```
arbrandes:
```suggestion
  "description": "Tutor frontend site",
```

tutormfe/templates/mfe/build/mfe/frontend-site/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/openedx/frontend-template-site.git"
+  },
```
arbrandes: You can drop a bunch of things from this package.json, since it's not going to be published.  Not even `name` is needed.  Let's just keep the bare minimum.

tutormfe/templates/mfe/build/mfe/frontend-site/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "serve": "openedx serve"
+  },
+  "author": "Open edX",
+  "license": "AGPL-3.0",
+  "homepage": "https://github.com/openedx/frontend-template-site#readme",
+  "bugs": {
+    "url": "https://github.com/openedx/frontend-template-site/issues"
+  },
+  "dependencies": {
+    "@edx/brand": "npm:@openedx/brand-openedx@^1.2.3",
```
arbrandes: Not sure if this is going to work, but we should try it.  The less hard-coded versions we have here, the better.

```suggestion
    "@edx/brand": "npm:@openedx/brand-openedx@latest",
```

tutormfe/templates/mfe/build/mfe/frontend-site/package.json
```diff
@@ -0,0 +1,54 @@
...
+{%- for app_name, app_attrs in iter_frontend_apps() %}
+    "@openedx/frontend-app-{{ app_name }}": "^1.0.0-alpha || 0.0.0-dev",
```
arbrandes: I _think_ the version is going to have to be configurable per app as a string property, so that it can be configured by operators.  `app_npm_version`, or something like that to distinguish it from the `app_git_version`.

I'm also not sure what to do about `|| 0.0.0-dev` in particular.  If we leave this as a per-app string, it's up to the operator to configure correctly - which I guess is good enough.  They should know if their own app is semantically released or not, and what the miminum version is.

tutormfe/templates/mfe/build/mfe/frontend-site/package.json
```diff
@@ -0,0 +1,54 @@
...
+{%- endfor %}
+    "@openedx/frontend-base": "^1.0.0-alpha || 0.0.0-dev"
```
arbrandes: This might also need to be configurable.  The scenario I'm thinking of is an operator that has their own published fork of frontend-base and wants to alias it here.

tutormfe/templates/mfe/build/mfe/frontend-site/package.json
```diff
@@ -0,0 +1,54 @@
...
+  },
+  "devDependencies": {
+    "@edx/browserslist-config": "^1.5.0",
```
arbrandes:
```suggestion
    "@edx/browserslist-config": "latest",
```

...probably.

tutormfe/templates/mfe/build/mfe/frontend-site/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "turbo": "^2.8.16"
+  },
+  "peerDependencies": {
+    "@openedx/paragon": "^23",
```
arbrandes: Let's also allow the operator to use their own fork of Paragon.  ("$paragon_version").

tutormfe/templates/mfe/build/mfe/frontend-site/site.config.build.tsx
```diff
@@ -0,0 +1,50 @@
+import { EnvironmentTypes, SiteConfig, footerApp, headerApp, shellApp } from '@openedx/frontend-base';
+{%- for app_name, app_attrs in iter_frontend_apps() %}
+{%- set camel_case_name = app_name.split('-') | map('title') | list %}
+{%- set camel_case_name = camel_case_name[0].lower() + (camel_case_name[1:] | join('')) %}
```
arbrandes: Same as above.  Let's derive this from `appId`, which the operator has to configure in the app list.

tutormfe/templates/mfe/build/mfe/frontend-site/site.config.build.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { {{ camel_case_name }}App } from '@openedx/frontend-app-{{ app_name }}';
```
arbrandes: Too fragile, IMHO.  Let's have the operator explicitly set this.  How about if we had an `entrypoints` dictionary, per app?  With something like:

```
entrypoints: {
  "scss": "app.scss",
  "js": "LearnerDashboardApp"
}
```

tutormfe/templates/mfe/build/mfe/frontend-site/src/site.scss
```diff
@@ -0,0 +1,6 @@
...
+@use '@openedx/frontend-app-{{ app_name }}/app.scss' as {{ camel_case_name }};
```
arbrandes: We should also let the user specify their own app's SCSS filename, as I suggest in another comment below.

tutormfe/templates/mfe/build/mfe/frontend-site/site.config.build.tsx
```diff
@@ -0,0 +1,50 @@
...
+{%- endfor %}
+import homeApp from './src/homeApp';
+
+
+import './src/site.scss';
+
+const siteConfig: SiteConfig = {
+  siteId: 'frontend-template-site',
+  siteName: 'Frontend Template Site',
```
arbrandes: Needs to be Tutor-configurable.

tutormfe/templates/mfe/build/mfe/frontend-site/site.config.build.tsx
```diff
@@ -0,0 +1,50 @@
...
+{%- set camel_case_name = camel_case_name[0].lower() + (camel_case_name[1:] | join('')) %}
+import { {{ camel_case_name }}App } from '@openedx/frontend-app-{{ app_name }}';
+{%- endfor %}
+import homeApp from './src/homeApp';
+
+
+import './src/site.scss';
+
+const siteConfig: SiteConfig = {
+  siteId: 'frontend-template-site',
```
arbrandes:
```suggestion
  siteId: 'tutor-frontend-site',
```

tutormfe/templates/mfe/build/mfe/frontend-site/site.config.build.tsx
```diff
@@ -0,0 +1,50 @@
...
+  siteName: 'Frontend Template Site',
+  baseUrl: '{{ "https" if ENABLE_HTTPS else "http" }}://{{ MFE_HOST }}:8080',
+  lmsBaseUrl: '{{ "https" if ENABLE_HTTPS else "http" }}://{{ LMS_HOST }}:8000',
+  loginUrl: '{{ "https" if ENABLE_HTTPS else "http" }}://{{ LMS_HOST }}:8000/login',
+  logoutUrl: '{{ "https" if ENABLE_HTTPS else "http" }}://{{ LMS_HOST }}:8000/logout',
```
arbrandes, line 21: Double-checking: is this correct production configuration?

tutormfe/templates/mfe/build/mfe/frontend-site/site.config.build.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  environment: EnvironmentTypes.PRODUCTION,
+  apps: [
+    shellApp,
+    headerApp,
+    footerApp,
+{%- for app_name, app_attrs in iter_frontend_apps() %}
+{%- set camel_case_name = app_name.split('-') | map('title') | list %}
+{%- set camel_case_name = camel_case_name[0].lower() + (camel_case_name[1:] | join('')) %}
+    {{ camel_case_name }}App,
```
arbrandes: Let's use entrypoint configuration, as noted above.

tutormfe/templates/mfe/build/mfe/frontend-site/site.config.build.tsx
```diff
@@ -0,0 +1,50 @@
...
+    },
+    {
+      role: 'org.openedx.frontend.role.account',
+      url: '{{ "https" if ENABLE_HTTPS else "http" }}://{{ MFE_HOST }}/account/'
+    },
+    {
+      role: 'org.openedx.frontend.role.logout',
+      url: '{{ "https" if ENABLE_HTTPS else "http" }}://{{ LMS_HOST }}/logout'
+    },
+  ],
```
arbrandes, line 44: For now this default is fine, but as soon as we have frontend-base-toggleable Profile and Account MFEs, we're going to need some logic here.

However, this is definitely going to have to be configurable off the bat.  In other words, Tutor plugins must be able to add arbitrary external routes as per the operator's wishes.

tutormfe/templates/mfe/build/mfe/frontend-site/site.config.build.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  accessTokenCookieName: 'edx-jwt-cookie-header-payload',
```
arbrandes: Also needs to be configurable.

tutormfe/templates/mfe/build/mfe/frontend-site/site.config.dev.tsx
```diff
@@ -0,0 +1,47 @@
+import { EnvironmentTypes, SiteConfig, footerApp, headerApp, shellApp } from '@openedx/frontend-base';
+{%- for app_name, app_attrs in iter_frontend_apps() %}
+{%- set camel_case_name = app_name.split('-') | map('title') | list %}
+{%- set camel_case_name = camel_case_name[0].lower() + (camel_case_name[1:] | join('')) %}
```
arbrandes: The comments for site.config.build also apply here, for the most part.

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -82,6 +82,33 @@ RUN npm run build
 {{ patch("mfe-dockerfile-post-npm-build-{}".format(app_name)) }}
 {% endfor %}
 
+
+{%- for app_name, app in iter_frontend_sites() %}
```
arbrandes:
```suggestion
{%- for site_name, site in iter_frontend_sites() %}
```

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -82,6 +82,33 @@ RUN npm run build
...
+######## build of site-{{ app_name }}
+FROM base AS site-{{ app_name }}-common
+WORKDIR /openedx/app
+
+{% if app.get("repository") and app["repository"] == "local" %}
+COPY frontend-site/ /openedx/app
+{% else %}
+RUN git clone --depth 1 --branch {{ app.get("version", MFE_COMMON_VERSION) }} {{ app["repository"] }} .
```
arbrandes: I think we need to guard for app.repository being undefined.

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -82,6 +82,33 @@ RUN npm run build
 {{ patch("mfe-dockerfile-post-npm-build-{}".format(app_name)) }}
 {% endfor %}
 
+
+{%- for app_name, app in iter_frontend_sites() %}
+######## build of site-{{ app_name }}
+FROM base AS site-{{ app_name }}-common
+WORKDIR /openedx/app
```
arbrandes:
```suggestion
WORKDIR /openedx/site
```

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -82,6 +82,33 @@ RUN npm run build
...
+{% else %}
+RUN git clone --depth 1 --branch {{ app.get("version", MFE_COMMON_VERSION) }} {{ app["repository"] }} .
+{% endif %}
+
+WORKDIR /openedx/app/packages
+# Git clone all frontend apps that should be built
+{% for app_name, app in iter_frontend_apps() %}
+{% if app.get("repository") %}
+RUN git clone --depth 1 --branch {{ app.get("version", MFE_COMMON_VERSION) }} {{ app["repository"] }} frontend-app-{{ app_name }}
+{% endif %}
```
arbrandes: As you mentioned elsewhere, we're missing the volume mounts.

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -82,6 +82,33 @@ RUN npm run build
...
+{% endfor %}
+
+WORKDIR /openedx/app
+ARG NPM_REGISTRY={{ NPM_REGISTRY }}
+RUN --mount=type=cache,target=/root/.npm-sites,sharing=shared npm install --no-audit --no-fund --registry=$NPM_REGISTRY
+RUN npm run build:packages
```
arbrandes, line 135: We need to guard against a site not having any local checkouts; say, if every app is just pulled from NPM.  In that case, we don't build:packages at all.

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -82,6 +82,33 @@ RUN npm run build
 {{ patch("mfe-dockerfile-post-npm-build-{}".format(app_name)) }}
 {% endfor %}
 
+
+{%- for app_name, app in iter_frontend_sites() %}
+######## build of site-{{ app_name }}
+FROM base AS site-{{ app_name }}-common
+WORKDIR /openedx/app
+
+{% if app.get("repository") and app["repository"] == "local" %}
```
arbrandes: I think it would be better to check for a specific site ID, instead of overloading `site["repository"]`.  Something like `if site["id"] == "default"`.

tutormfe/plugin.py
```diff
@@ -217,6 +304,8 @@ def _mounted_mfe_image_management() -> None:
     tutor_hooks.Filters.CLI_DO_INIT_TASKS.add_item(("lms", task_file.read()))
 
 REPO_PREFIX = "frontend-app-"
+FRONTEND_SITE_PREFIX = "frontend-site-"
+# TODO: figure out the whole mount thing not only for sites but also for the possible apps under it
```
arbrandes: Yes. :)

tutormfe/templates/mfe/build/mfe/frontend-site/src/homeApp.ts
```diff
@@ -0,0 +1,18 @@
+import { App, getUrlByRouteRole } from '@openedx/frontend-base';
+import { redirect } from 'react-router';
+
+const homeApp: App = {
+  appId: 'home',
+  routes: [{
+    path: '/',
+    loader: () => {
+      const dashboardUrl = getUrlByRouteRole('org.openedx.frontend.role.dashboard');
```
arbrandes: I think the home role will have to be configurable.

Together with a configurable `externalRoutes` (and the list of apps, obviously), this would let the user redirect anywhere they want.

tutormfe/templates/mfe/apps/mfe/Caddyfile
```diff
@@ -26,15 +26,34 @@
...
     {% endif %}
 
-    {% for app_name, app in iter_mfes() %}
+    {% for app_name, app in iter_paths() %}
     @mfe_{{ app_name }} {
         path /{{ app_name }} /{{ app_name }}/*
     }
     handle @mfe_{{ app_name }} {
         uri strip_prefix /{{ app_name }}
+        {%- if is_frontend_app(app_name) %}
```
arbrandes: If in `iter_paths()` we're calling `iter_frontend_apps()` and `iter_mfes()`, and here there's not a lot in common between the two, maybe the unification is not worth it.  I think we should just separate the loops in the Caddyfile and ditch `iter_paths()` entirely.

It also has the benefit of clarity, because then we could have `@app_{{ app_name }}` and `@mfe_{{ mfe_name }}` instead of `@mfe_{{ app name }}`.

arbrandes (changes requested):
Great first stab, thank you!

I do have a lot of comments, though.  Apologies in advance.

Also, it would be good to have an addition to the README explaining not only how to use this stuff, but how it works.

tutormfe/templates/mfe/apps/mfe/Caddyfile
holaontiveros: iter_paths doesn't call the iter functions but the "get" ones, but I think that's what you meant, the thing here it's that we are not looking for the whole list of those, because if that was the case then we could just iterate over both of them, we are just looking for the list of unique paths specifically for cases like the instructor dashboard

`get_mfes` returns the list of all mfes which at this point we can consider that we need all those paths

`get_frontend_apps`returns the list of all the frontend_apps configured which normally should be anything we want to route as a frontend_app for example learning, authn and instructor

the thing now it's that given that everything it's inside the same Caddy instance any request to an MFE always reaches the same place and to return the right thing we need a non-repeated list of those paths which is what iter_paths does

I do 100% agree with the @apps part tho it would make it way more clear I'll be changing that right away

Example of how it looks after that change:

```Caddyfile
 @mfe_gradebook {
        path /gradebook /gradebook/*
    }
    handle @mfe_gradebook {
        uri strip_prefix /gradebook
        root * /openedx/dist/gradebook
        try_files /{path} /index.html
        file_server
    }
    
    @app_learner-dashboard {
        path /learner-dashboard /learner-dashboard/*
    }
    handle @app_learner-dashboard {
        uri strip_prefix /learner-dashboard
        root * /openedx/dist/site-default
        try_files /{path} /index.html
        file_server
    }
```

tutormfe/templates/mfe/apps/mfe/Caddyfile
holaontiveros: separating the mfes and apps completely would make a bit more readable the Caddyfile (which normally is not read by anyone) but the rest of the code a bit harder to follow to my opinion

For example we still need to iter_mfes in other places, so we would need iter_mfes_paths to filter the mfes that are enabled as frontend_apps and it would be exclusive for caddy, then iter_frontend_apps which could just be the normal iter for frontend_apps

either way we end up with an extra function that it's only used for the caddyfile but I feel that with the `@mfe_app_name` and `@frontend_app_name` should be explicit enough to understand what's happening in each one of those

what do you think @brian-smith-tcril ?

the sites can be at whichever part, for now I had here at the end because it's easy to find it there

tutormfe/templates/mfe/build/mfe/frontend-site/src/homeApp.ts
holaontiveros: The externalRoutes we could store it on the site config, so it's overridable through that same config and if we start having those there probably we can do the same for the role here I'll take a stab at it

tutormfe/templates/mfe/apps/mfe/Caddyfile
arbrandes: You're right, I forgot about the uniqueness thing!  Never mind, then.

tutormfe/templates/mfe/build/mfe/Dockerfile
holaontiveros: It was like this probably  because before there was the "if present" on the build stuff, is there something equivalent on turbo? or do I need do the custom thin

tutormfe/templates/mfe/build/mfe/Dockerfile
holaontiveros: default can be overriden by the operator and if that happens we need to know if it's a real repository or it's the local files

tutormfe/templates/mfe/build/mfe/Dockerfile
holaontiveros: it would only be undefined if they are replacing the default or adding a new site and missed adding it and at that point we want the build to fail right?

tutormfe/templates/mfe/build/mfe/frontend-site/site.config.build.tsx
holaontiveros: no it's not, but while I get to use one for prod and one for dev, having this here makes testing on dev way easier

I'm getting there c:

brian-smith-tcril:
Took a quick look, most things appear to make sense at first glance, the one thing that stood out to me was:
* What's the purpose of `"styles": ["app.scss"]` in `appEntryPoints`?

Looking into the code and some comments from @arbrandes it started to make sense (it's so we can handle what https://github.com/openedx/frontend-template-site/blob/main/src/site.scss does).

So this really becomes more of a `frontend-base`/`frontend-template-site` question than a `tutor-mfe` question: why do site operators using `frontend-template-site` need to manually add their apps to both `site.config` and `site.scss`?

holaontiveros:
@brian-smith-tcril  I'm guessing because we don't import the css at the `App` level but at the site.config level, and given that the template-site it's a different site.config from the original dev one from the app then we need to have the imports of all the things somwhere...

Now why we do it like that and not somewhere else may be related with a decision on how we are using webpack maybe? @arbrandes

arbrandes:
> why do site operators using frontend-template-site need to manually add their apps to both site.config and site.scss

The simple answer: app SCSS must be imported somewhere by the consuming application (in our case, the site) because apps are not distributed with pre-compiled CSS.

If y'all can think of a better way to handle this, I'm all ears!

arbrandes:
Ah, turns out we actually documented this [in ADR 8](https://github.com/openedx/frontend-base/blob/main/docs/decisions/0008-stylesheet-import-in-site-config.md), way-back-when.  It's a webpack quirk.

arbrandes:
@holaontiveros, thanks a lot for this!  I squashed this PR into [a single commit](https://github.com/overhangio/tutor-mfe/pull/285/changes/2739ecda65d1d87373133627c5da17f6b94075d4) in [a new upstream frontend-base branch](https://github.com/overhangio/tutor-mfe/tree/frontend-base) (and [associated PR](https://github.com/overhangio/tutor-mfe/pull/285)) we can use for coordination.  While I'm closing this, consider it merged! :)

Note, however, that I already made significant changes to the mechanism in order to simplify it and bring it more in line with how tutor (and tutor-mfe) usually do things (more sources of truth inside templates, instead of configuration variables).  I also took a first stab at adding dev mount support (for sites only, so far - still working on apps).

PR #285, opened 2026-04-03
feat: add frontend-base support

arbrandes:
### Description

**Parent issue:**  https://github.com/openedx/frontend-base/issues/133

**Sandbox:** https://frontend-base.openedx.io/

This PR adds support for [frontend-base](https://github.com/openedx/frontend-base), a unified framework that replaces `frontend-build`, `frontend-platform`, `frontend-plugin-framework`, `frontend-component-header`, and `frontend-component-footer`. It enables frontend apps to be loaded as direct plugins within a single shell application, rather than as separate, independently deployed SPAs.

When frontend apps are configured, the plugin builds a frontend-base site that bundles them together into this shell. The site is served by the same `mfe` container alongside legacy MFEs, with Caddy routing requests appropriately.

Four frontend apps are shipped here.  Two new ones, enabled by default: `instructor-dashboard` and `notifications`.  And two replacements for legacy MFEs: `authn`, `learner-dashboard`.  These two are disabled by default.  (Because we want to enable all new-style apps at the same time once all MFEs are converted, to minimize operator friction.)

### Implementation

The build infrastructure uses an npm workspaces monorepo with some usual suspects in the Dockerfile (`site-common`, `site-prod`, `mfe-dev`). Frontend apps not configured to be pulled from NPM (such as when using a git repository or mounted for development) are installed and built as workspace packages. If one matches a legacy MFE, both are still built so that enabling/disabling apps can be done without an image rebuild, but the legacy MFE is effectively disabled for the end user.

Runtime configuration is fetched from the LMS at `/api/frontend_site_config/v1/`, populated from a new `FRONTEND_SITE_CONFIG` dictionary in the LMS settings. For backward compatibility, the endpoint also converts existing `MFE_CONFIG` and `MFE_CONFIG_OVERRIDES` to frontend-base's SiteConfig structure, at a lower precedence.

In dev mode, mounting a `frontend-site` directory or any `frontend-app-*` repository starts a single `mfe-dev` service for all apps, with hot-reload support.  If the appropriate `tutor mount add`s are run, individual frontend app repos are bind-mounted as npm workspace packages into both the `mfe` and `mfe-dev` containers.

Frontend apps (`FRONTEND_APPS`) are introduced as a new hook alongside the existing `MFE_APPS`, with corresponding template helpers, LMS settings patches, and a full set of `mfe-site-*` template patches for customization.

New Tutor configuration variables: `MFE_SITE_PORT`, `MFE_SITE_REPOSITORY`, and `MFE_SITE_VERSION`.

### Upstream blockers

- [x] https://github.com/openedx/frontend-base/pull/209/
- [x] https://github.com/openedx/openedx-platform/pull/38061
- [x] https://github.com/openedx/frontend-base/pull/227

### Testing

This PR requires an up-to-date deployment of [tutor@main](https://github.com/overhangio/tutor/tree/main) (in particular, you'll need https://github.com/openedx/openedx-platform/commit/57914c56a569b224b0074bb0c291a58df9b26788 or newer). Make sure your Tutor installation is current before proceeding.

Install this PR's version of tutor-mfe:

```
pip install "git+https://github.com/overhangio/tutor-mfe.git@frontend-base"
```

Then save the configuration and rebuild images. Because the `mfe` and `mfe-dev` images both need rebuilding, a `tutor dev launch` is recommended:

```
tutor config save
tutor dev launch
```

**Baseline: legacy MFEs should still work.** With authn and learner-dashboard disabled (the default), everything should behave exactly as before. All previous existing MFEs should load and function normally, including in development mode. There should be no detectable changes in behavior.

**Enabling frontend apps.** Next, enable the two core frontend apps that ship disabled (instructor-dashboard and notifications are enabled by default, already). Create a Tutor plugin (e.g. `"$(tutor plugins printroot)/enable_frontend_apps.py"`) with the following content:

```python

@FRONTEND_APPS.add()
def _enable_core_apps(apps):
    apps["authn"]["enabled"] = True
    apps["learner-dashboard"]["enabled"] = True
    return apps
```

Then enable the plugin and restart (no image rebuild needed):

```
tutor plugins enable enable_frontend_apps
tutor config save
tutor dev stop && tutor dev start -d
```

The frontend-base apps should then be accessible in the browser after being redirected from `http://local.openedx.io:8000` to `http://apps.local.openedx.io:8080`.  If you're logged in, you should be taken to the dashboard directly at `http://apps.local.openedx.io:8080/learner-dashboard/`.  If not, you should be taken to `http://apps.local.openedx.io:8080/authn/login`.

**Note:** While the focus of this PR is on the frontend-base integration infrastructure, not on the functionality of Authn or Learner Dashboard themselves, any bugs encountered in those apps should still be noted.

**Development modes.** Try mounting individual frontend app repositories and also the whole frontend site for development. Note the behavior difference depending on whether a given app is enabled: when an app is enabled, mounting it provides hot-reload development of that app within the frontend-base site; when it is not enabled, mounting it starts it as a standalone MFE instead (the legacy behavior). See the ["Frontend-base site development" section of the README](https://github.com/overhangio/tutor-mfe/blob/9eaec0e41f63540a39eeee0a9199a65f2fdaf5b7/README.rst#L938-L965) for details.

### TODO

- [x] POC implementation (pending [upstream blocker](https://github.com/openedx/frontend-base/pull/209/))
- [x] Use new runtime configuration endpoint (pending [the other upstream blocker](https://github.com/openedx/openedx-platform/pull/38061))
- [x] Documentation
- [x] Wait for upstream blockers to land
- [x] Add package-lock.json (with a sane way to update it later)
- [x] Testing

### LLM usage notice


tutormfe/plugin.py
```diff
@@ -82,6 +93,19 @@
     },
 }
 
+CORE_FRONTEND_APPS: dict[str, FRONTEND_APP_ATTRS_TYPE] = {
+    "authn": {
+        "npm_package": "@openedx/frontend-app-authn",
+        "npm_version": "^1.0.0-alpha || 0.0.0-dev",
+        "enabled": True,
```
arbrandes: It is the consensus in Axim engineering that these should _not_ be enabled by default for Verawood, primarily because users would have to engage with two different slot configuration APIs: one for Authn and Learner Dash, and another for all the other MFEs.

To get some usage data, we should instead:

* Try to publicize how easy it is to enable these two apps on and test "the future"
* Since this PR is expected to land, we can try to add some frontend-base testing to the Verawood testing plan, which should be made possible if we provide a sandbox.

arbrandes:
Upstream blockers have landed, so this is now ready for review.

tutormfe/patches/local-docker-compose-dev-services
```diff
@@ -19,8 +20,25 @@
...
+        {%- endfor %}
+    restart: unless-stopped
+    depends_on:
+        - lms
+    environment:
+        - "PORT={{ MFE_SITE_PORT }}"
+{%- endif %}
 
-{% if mfe_data.unmounted|length > 0 or MFE_HOST_EXTRA_FILES %}
+{% if mfe_data.unmounted|length > 0 or not site_mounts or MFE_HOST_EXTRA_FILES %}
```
brian-smith-tcril, line 44: Curious about this logic. I think the main thing that is making it seem confusing to me is that `site_mounts` (plural) implies we might have multiple sites. In the case where multiple sites exist, one is mounted and one isn't, then `site_mounts` would be truthy.

I see 2 possible solutions here:
1. Rename `get_site_mounts` and `site_mounts` to `get_site_mount` and `site_mount` respectively. That would clarify the limitation of "we only support one site."
2. Support multiple sites. One possible way to do this would be to follow the `MFEMountData` class pattern for a `SiteMountData` class.

arbrandes: > site_mounts (plural) implies we might have multiple sites

Not necessarily, if you take it to mean: "all of the site-related mounts that go into the `mfe-dev` service".  This includes both the site itself, and/or the apps.  So the name is fine, I think.

On the suggestions:

1. Doesn't work, because there are really multiple mounts possible.
2. That is explicitly a non-goal of this implementation: there is no scenario I can imagine where multiple sites is a desirable tutor-mfe feature.  Users can already override the full site config if they want - I can see why somebody would want to do that - but multiple sites negates most (or at least, many) of the benefits of frontend-base.

brian-smith-tcril: Gotcha. So with

```py
        if is_frontend_app_enabled(app_name):
            mounts.append(("mfe", f"frontend-app-{app_name}-src"))
            mounts.append(("mfe-dev", f"frontend-app-{app_name}-src"))
```

the `frontend-app-*` mounts are "site mounts."

Seems like a classic "names are hard" problem. When I first saw `get_site_mounts` I read it as "get mounts for all the sites" instead of "get all the mounts for the one site."

arbrandes: I'm actually looking into what we can do, here.  Be back in a bit. :)

arbrandes: Ok, I don't have a better idea for the name (what I commented earlier holds), but you did help clarify the implementation:

78618aed

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -22,66 +22,130 @@ ENV PATH=/openedx/app/node_modules/.bin:${PATH}
...
+######## mfe (dev)
+FROM site-common AS mfe-dev
+ENV NODE_ENV=development
+EXPOSE {{ MFE_SITE_PORT }}
+CMD ["/bin/bash", "-c", "npm run dev:packages"]
+
+######## site (production)
+FROM site-common AS site-prod
+{{ patch("mfe-dockerfile-pre-npm-build-site") }}
+RUN npm run build:packages
```
holaontiveros, line 154: Are we not running the site build with the published packages at all?

holaontiveros: never mind, I did see what you did in the top by checking the repository, but one comment you had it was that we should ward agains doing this build when not required

README.rst
```diff
@@ -638,6 +638,300 @@ This works for custom MFEs, as well. For example, if you added your own MFE name
...
+Enabling or disabling existing apps does not require rebuilding tutor-mfe images: after a `tutor config save`, only a Tutor restart is required.
+
+To add a custom frontend app (which does require rebuilding), the Tutor plugin would be:
+
+.. code-block:: python
+
+    from tutormfe.hooks import FRONTEND_APPS
+
+    from tutor import hooks
+
```
holaontiveros, line 786: How do we use a custom version of frontend-base?

tutormfe/templates/mfe/build/mfe/Dockerfile
arbrandes: `build:packages` falls back gracefully: if there are no packages under `site/packages/`, it does nothing.  In other words, if there are no apps with configured repositories or nothing is mounted, the build uses only NPM.

arbrandes: (Yes, I only confirmed that later.)

README.rst
arbrandes: You either have to use your own site (either via setting MFE_SITE_REPOSITORY or by `tutor mounts add /path/to/frontend-site`), or you add a [`post-npm-install-site`](https://github.com/overhangio/tutor-mfe/blob/frontend-base/README.rst#L1284-L1284) patch.

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -22,66 +22,132 @@ ENV PATH=/openedx/app/node_modules/.bin:${PATH}
...
+{%- for app_name, app in iter_frontend_apps() %}
+COPY --from=frontend-app-{{ app_name }}-src / /openedx/site/packages/frontend-app-{{ app_name }}
+{%- endfor %}
+
+# Copy only package.json first so that npm install is cached independently of source changes
+COPY --from=site-src /package.json .
+
+ARG NPM_REGISTRY={{ NPM_REGISTRY }}
+{{ patch("mfe-dockerfile-pre-npm-install-site") }}
+RUN --mount=type=cache,target=/root/.npm-sites,sharing=shared npm install --no-audit --no-fund --registry=$NPM_REGISTRY
```
arbrandes, line 137: On the main branch there is no `package-lock.json` so that we always get the latest versions of all dependencies (in particular, the app packages) on every build.

Once this gets to a release branch, however, a package-lock should definitely be added so that a) builds are reproducible, and b) when dependencies are explicitly `npm update`d and checked-in, this will invalidate the Docker cache properly.

ahmed-arb:
Great work on this PR. Legacy MFEs behave as before, and once logged in the learner dashboard and authn apps work well within the shell.

I did run into one issue and wanted to check if I'm missing a step.

**Setup**: tutor 21.0.3-main, edx-platform pinned to 57914c56a5, frontend apps enabled (authn + learner-dashboard) via the plugin from the PR description.

**Issue**: Visiting http://apps.local.openedx.io:8080/ shows a 404. The frontend-base shell loads but has no route for /. After login, authn navigates client-side to next=/, which hits the same 404. Manually typing http://local.openedx.io:8000 or refreshing after login does land on /learner-dashboard/ correctly.                                                                                                                 
 
It looks like in dev mode the outer Caddyfile's redir / is bypassed (port 8080 maps directly to the MFE container's internal Caddy), and the built bundle doesn't seem to include org.openedx.frontend.role.home, so the shell router's automatic / to /learner-dashboard/ redirect doesn't kick in.

Am I missing a configuration step, or is this a known limitation of the current build?

Screenshot 2026-04-16 at 7 42 22 PM

arbrandes:
Thanks for testing this, @ahmed-arb!

> Am I missing a configuration step, or is this a known limitation of the current build?

I suspect your build was cached and did not get the latest version of frontend-base, which fixes this issue.  Please try:

```bash
tutor images build mfe --docker-arg --no-cache-filter=site-common
```

Confirm you're getting the latest frontend-base (1.0.0-alpha.27 or greater) by looking at the log:

```bash
=> [site-prod 2/2] RUN npm run build                                                                                                                                                                                                                    12.5s
 => => # > build
 => => # > openedx build
 => => # ████████████████████████████████
 => => # █ Open edX CLI v1.0.0-alpha.27 █
 => => # ████████████████████████████████
```

Then launch as you usually would.  If this still happens, let me know.

ahmed-arb:
Rebuilding with `--no-cache-filter=site-common` fixed the issue. Thanks for the pointer. The stale cache was indeed pulling an older `frontend-base`.

One related question came up while testing:

**Theming:** We maintain [[tutor-indigo](https://github.com/overhangio/tutor-indigo)](https://github.com/overhangio/tutor-indigo), which currently overrides `@edx/frontend-component-header` and `@edx/brand` via npm for individual MFEs (authn, learner-dashboard, etc.). Since `frontend-base` bundles apps into a single shell, those per-MFE npm overrides likely won't apply the same way. Is there a recommended approach for theming and branding the `frontend-base` shell?

arbrandes:
> The stale cache was indeed pulling an older frontend-base

As we discussed on Slack, the best way to prevent this going forward is going to be adding a package-lock.json.  That way builds are reproducible, and if anything new is merged, the Docker cache will be busted at the right spot automatically.  The down side is having to maintain that file (updating it whenever new dependencies come out), but it's worth it.

> Is there a recommended approach for theming and branding the frontend-base shell?

Header and footer customization is now handled via frontend-base "apps".  There's not much user-facing documentation for that, yet, but [this new section in the README](https://github.com/overhangio/tutor-mfe/pull/285/changes#diff-7b3ed02bc73dc06b7db906cf97aa91dec2b2eb21f2d92bc5caa761df5bbc168fR852) in this PR is as good a start as any.  TLDR: you would create a "frontend-app-indigo" package (name is arbitrary) following a very similar structure to the sample [devHeader app](https://github.com/openedx/frontend-base/tree/main/shell/dev/devHeader) in frontend-base itself.  The key is understanding [the slots array](https://github.com/openedx/frontend-base/blob/main/shell/dev/devHeader/app.tsx#L11).

Styling is still handled by @edx/brand the same way, but the `npm install` would go in the site's post-install patch.

I'll be glad to help out with this however I can.

arbrandes:
@ahmed-arb, I added a package-lock, as well as tooling to make updating it easier.  And because Renovate cannot be used (because package.json is a Jinja template), I added a (daily?) Github workflow to do the job.

arbrandes:
This is blocked on https://github.com/openedx/frontend-base/pull/227.  Until then, the build fails:

```
ERROR in ./src/utils.ts:1:30
23.65 TS2305: Module '"@openedx/frontend-base"' has no exported member 'ExternalScriptLoaderClass'.
23.65   > 1 | import { App, ExternalRoute, ExternalScriptLoaderClass, SiteConfig, SlotOperation } from '@openedx/frontend-base';
```

arbrandes:
We're unblocked: https://github.com/openedx/frontend-base/pull/227 has merged.

.github/workflows/update-site-lockfile.yml
brian-smith-tcril, line 1: Overall this workflow makes sense to me, but I'm slightly concerned that it won't catch build failures with the updated lockfile.

I know in theory things should stay compatible with our pinning strategy, but I've run into issues where dependencies unintentionally release a breaking change as minor (most recently https://github.com/openedx/paragon/issues/4218 / https://github.com/openedx/paragon/pull/4257)

With renovate PRs this gets caught by CI, but I don't see anything in `.github/workflows/test.yml` or the `test` target in the `Makefile` that would catch things like that.

arbrandes: Agreed, but the problem is that tutor mfe has never run CI against the actual build, probably because it's too expensive.  Probably good for a follow-up conversation with the other maintainers.

arbrandes: (Or it's just that this already runs as part of the Docker build process somewhere else not visible from here - which I think is the likely answer.)

tutormfe/templates/mfe/build/mfe/Dockerfile
```diff
@@ -22,74 +23,172 @@ ENV PATH=/openedx/app/node_modules/.bin:${PATH}
...
+# No need to build the frontend site if there are no frontend apps.
+{%- if get_frontend_apps() %}
+####################### site
+######## site (repo)
+FROM base AS site-repo
+WORKDIR /openedx/site
+
+{%- if MFE_SITE_REPOSITORY %}
+ADD --keep-git-dir=true {{ MFE_SITE_REPOSITORY }}#{{ MFE_SITE_VERSION }} .
+{%- else %}
```
Faraz32123, line 115: can we add `MFE_SITE_VERSION` in if condition as well.

tutormfe/templates/mfe/build/mfe/site/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "serve": "openedx serve",
+    "translations:pull": "openedx translations:pull"
+  },
+  "atlasTranslations": {
+    "dependencies": [
+{%- for app_name, app in iter_frontend_apps() %}
+{%- if app.get("npm_package") %}
+      "{{ app['npm_package'] }}"{{ "," if not loop.last }}
+{%- endif %}
+{%- endfor %}
```
Faraz32123: This might break if `loop.last` does not have a npm_package. The second-to-last app that had the npm package will have a trailing comma, will result in invalid JSON.

Faraz32123 (changes requested):

tutormfe/templates/mfe/build/mfe/Dockerfile
arbrandes: Good catch.  Addressed!

tutormfe/templates/mfe/build/mfe/site/package.json
arbrandes: Good catch.  This loop was not only dumb, it was wrong (it was missing frontend-base itself).  Fixed.

Faraz32123 (approved):
LGTM!

PR #286, opened 2026-04-06
fix(temporary): use patched frontend-base for relative URL fix

arbrandes:
### Description

This temporarily replaces the `@openedx/frontend-base` dependency with a patched version (`@arbrandes/frontend-base`) that fixes a relative URL resolution issue.

This is meant to be reverted once the fix is merged upstream in `frontend-base` itself.


arbrandes:
The fix has landed upstream.

PR #287, opened 2026-04-15
Chore: drop compatibility with Python 3.9 and ensure compatibility with Python 3.14

Syed-Ali-Abbas-568:
- Drop Python 3.9 (EOL) — set requires-python to >=3.10
- Add Python 3.13 and 3.14 classifiers in pyproject.toml
- Update CI matrix from ["3.9", "3.12"] to ["3.10", "3.14"]

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.12"]
+        python-version: ["3.10", "3.14"]
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.12"]
+        python-version: ["3.10", "3.14"]
```
arbrandes, line 14: Shouldn't this be:


```suggestion
        python-version: ["3.12", "3.14"]
```

In other words, dropping the oldest, adding a new one?

Syed-Ali-Abbas-568: Thanks for the review! I wanted to flag that the original [issue #1306](https://github.com/overhangio/tutor/issues/1306#issue-3662238499) explicitly specifies upgrading CI from 3.9/3.12 to 3.10/3.14, so this change follows that spec directly. That said, happy to update if the requirements have since changed,  just let me know

arbrandes: Ok!

PR #289, opened 2026-04-15
feat: add INJECTED_SCRIPTS hook for MFE script injection

sagrodat:
Creates a parallel pipeline to the `PLUGIN_SLOTS` pattern specifically for injecting raw JavaScript classes into `env.config.jsx`. Prevents injected scripts from clobbering default external scripts or React components by injecting before the `PLUGIN_SLOTS` mechanism which uses FPF (frontend-plugin-framework) that this slot mechanism does not need.

### How to test this PR
To verify the injection pipeline works, you can drop this simple test plugin into your local environment.

```python
from tutormfe.hooks import INJECTED_SCRIPTS

test_content = """
class TestScriptLoader {
  constructor(data) { this.data = data; }
  loadScript() {
    console.log("Test Script injected successfully from Tutor!");
  }
}
"""

# Injecting into the 'authn' MFE for testing
INJECTED_SCRIPTS.add_item(
  ("authn", "test_script", test_content)
)
```
Install, enable and save config it:
```
tutor plugins install test_script.py
tutor plugins enable test_script
tutor config save
```
Check the generated file at `env/plugins/mfe/build/mfe/env.config.jsx` to see the script appended, or copy to the mounted MFE's directory and run the MFE to see the console in dev tools.

### DIFF
--- a/tutormfe/hooks.py
+++ b/tutormfe/hooks.py
@@ -15,3 +15,5 @@ MFE_ATTRS_TYPE = t.Dict[t.Literal["repository", "port", "version"], t.Union["str
 MFE_APPS: Filter[dict[str, MFE_ATTRS_TYPE], []] = Filter()
 
 PLUGIN_SLOTS: Filter[list[tuple[str, str, str]], []] = Filter()
+
+INJECTED_SCRIPTS: Filter[list[tuple[str, str, str]], []] = Filter()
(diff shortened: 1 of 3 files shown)


arbrandes:
Superseded by https://github.com/overhangio/tutor-mfe/pull/290.

PR #292, opened 2026-04-20
feat: add CORE_PLUGINS hook and bundle the notifications tray

arbrandes:
### Description

Ports `tutor-contrib-platform-notifications` into tutor-mfe so that enabling the plugin gives operators the notifications tray out of the box, with no extra plugin to install.

To support this cleanly, it introduces a `CORE_PLUGINS` filter in `tutormfe.hooks`: a registry of bundled frontend plugin packages that ship enabled by default. Operators can disable any core plugin by popping its entry from the filter, symmetrically to how `MFE_APPS` works today. Notifications is the first (and currently only) core plugin.

### Testing

With this branch installed and tutor-mfe enabled, run `tutor config save` and inspect the rendered environment. In `$(tutor config printroot)/env/plugins/mfe/build/mfe/Dockerfile` each MFE stage should contain a `RUN ... npm install ... @edx/frontend-plugin-notifications@^2.0.3` line just before the `mfe-dockerfile-post-npm-install` patch call. In `$(tutor config printroot)/env/plugins/mfe/build/mfe/env.config.jsx` there should be a `const { NotificationsTray } = await import('@edx/frontend-plugin-notifications');` line inside the `try {}` block. In `$(tutor config printroot)/env/apps/openedx/settings/lms/production.py`, both `MFE_CONFIG["SHOW_EMAIL_CHANNEL"]` and `MFE_CONFIG["SHOW_PUSH_CHANNEL"]` should be set, and `NOTIFICATIONS_DEFAULT_FROM_EMAIL` should be defined via `ENV_TOKENS`.

To verify the disable path, create a file at `$(tutor plugins printroot)/disable_notifications.py` containing:

    from tutormfe.hooks import CORE_PLUGINS

    @CORE_PLUGINS.add()
    def _disable_notifications(plugins):
        plugins.pop("notifications", None)
        return plugins

Then run `tutor plugins enable disable_notifications && tutor config save`. The rendered Dockerfile, `env.config.jsx`, LMS production settings, and `lms-env` should no longer reference `frontend-plugin-notifications`, `NotificationsTray`, `SHOW_EMAIL_CHANNEL`, or `NOTIFICATIONS_DEFAULT_FROM_EMAIL`. Running `tutor plugins disable disable_notifications && tutor config save` brings them all back.

### LLM usage notice

### DIFF
--- a/tutormfe/templates/mfe/build/mfe/env.config.jsx
+++ b/tutormfe/templates/mfe/build/mfe/env.config.jsx
@@ -25,8 +25,31 @@ async function setConfig () {
      * needs to be inside the `try{}` block.
      */
     const { DIRECT_PLUGIN, PLUGIN_OPERATIONS } = await import('@openedx/frontend-plugin-framework');
+    {%- if is_core_plugin_enabled("notifications") %}
+    const { NotificationsTray } = await import('@edx/frontend-plugin-notifications');
+    {%- endif %}
 
     {{- patch("mfe-env-config-runtime-definitions") }}
+
+    {%- if is_core_plugin_enabled("notifications") %}
+    {%- for slot_name in [
+      "org.openedx.frontend.layout.header_desktop_secondary_menu.v1",
+      "org.openedx.frontend.layout.header_learning_help.v1",
+      "org.openedx.frontend.layout.studio_header_search_button_slot.v1",
+    ] %}
+    addPlugins(config, '{{ slot_name }}', [
+      {
+        op: PLUGIN_OPERATIONS.Insert,
+        widget: {
+          id: 'notification-drawer-widget',
+          priority: 10,
+          type: DIRECT_PLUGIN,
+          RenderWidget: NotificationsTray,
+        },
+      },
+    ]);
+    {%- endfor %}
+    {%- endif %}
 
     {%- for slot_name, plugin_config in iter_plugin_slots("all") %}
     addPlugins(config, '{{ slot_name }}', [{{ plugin_config }}]);
--- a/README.rst
+++ b/README.rst
@@ -18,6 +18,10 @@ In addition, this plugin comes with a few MFEs which are enabled by default:
 - `Profile <https://github.com/openedx/frontend-app-profile/>`__
 - `Catalog <https://github.com/openedx/frontend-app-catalog/>`__
 
+In addition, this plugin bundles a number of "core plugins": frontend plugin packages injected into the MFEs above via the plugin slot framework. The following core plugins are enabled by default:
+
+- `Notifications <https://github.com/openedx/frontend-plugin-notifications/>`__ (a notifications tray in the MFE headers)
+
 Instructions for using each of these MFEs are given below.
 
 Installation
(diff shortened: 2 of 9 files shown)


tutormfe/templates/mfe/build/mfe/env.config.jsx
```diff
@@ -25,9 +25,32 @@ async function setConfig () {
      * needs to be inside the `try{}` block.
      */
     const { DIRECT_PLUGIN, PLUGIN_OPERATIONS } = await import('@openedx/frontend-plugin-framework');
+    {%- if is_core_plugin_enabled("notifications") %}
+    const { NotificationsTray } = await import('@edx/frontend-plugin-notifications');
+    {%- endif %}
```
Faraz32123, line 30: can we add this to `mfe-env-config-runtime-definitions` patch again like it was in [tutor-contrib-platform-notifications](https://github.com/openedx/tutor-contrib-platform-notifications/blob/main/tutorplatformnotifications/patches/mfe-env-config-runtime-definitions) instead of filling in `env.config.jsx`?

tutormfe/templates/mfe/build/mfe/env.config.jsx
```diff
@@ -25,9 +25,32 @@ async function setConfig () {
...
+        widget: {
+          id: 'notification-drawer-widget',
+          priority: 10,
+          type: DIRECT_PLUGIN,
+          RenderWidget: NotificationsTray,
+        },
+      },
+    ]);
+    {%- endfor %}
+    {%- endif %}
```
Faraz32123, line 52: can we also move this to `plugin.py` file under notifications section? As I was thinking we might end up adding plugin slots directly to env.config.jsx in future? Although its also fine if we don't do it.

Faraz32123 (approved):

Faraz32123:
@arbrandes Do we need this PR? as most of these changes also exists in frontend-base [PR](https://github.com/overhangio/tutor-mfe/pull/285) e.g. `NOTIFICATIONS_DEFAULT_FROM_EMAIL` exists in both, hooks(`CORE_PLUGINS` & `CORE_PLUGIN_ATTRS_TYPE` also seems redundant & exists with different names in frontend-base PR) etc.

tutormfe/templates/mfe/build/mfe/env.config.jsx
arbrandes: We could, but not cleanly:  the patches tutor-mfe defines (such as `mfe-env-config-runtime-definitions`) are meant for other plugins to consume.  That's what the `templates` are for: we put the stuff we own in there, and patches are for external use.

tutormfe/templates/mfe/build/mfe/env.config.jsx
arbrandes: I actually tried, but it doesn't work well: we're consuming the hook (PLUGIN_SLOTS) we're defining ourselves, so there are timing/race conditions that result in the core plugin not being picked up here.

I know it looks funny, but defining this directly in the template is the right place.

arbrandes:
@Faraz32123, we do need this PR.  It looks similar to frontend-base, but frontend-plugin-notifications and frontend-app-notifications are two separate things: the first is for legacy MFEs (which this PR takes care of), the second for frontend-base apps.

Once this PR merges, I'll have to rebase the frontend-base one so the two conditions are checked (is_core_plugin_enabled("notifications") and is_frontend_app_enabled("notifications"), for the backend stuff.

(Yes, this looks pretty redundant, but until all MFEs are converted to apps, we're going to have to live with some duplication.)

tutormfe/templates/mfe/build/mfe/env.config.jsx
Faraz32123: > there are timing/race conditions that result in the core plugin not being picked up here.
> 
> I know it looks funny, but defining this directly in the template is the right place.

makes sense,

Faraz32123 (approved):
LGTM!

PR #294, opened 2026-04-22
Revert "feat: add CORE_PLUGINS hook and bundle the notifications tray"

arbrandes:
### Description

Reverts 84368d9 ("feat: add CORE_PLUGINS hook and bundle the notifications tray"), which introduced a `CORE_PLUGINS` filter and bundled the notifications tray as the first core plugin.

The rationale for this revert is that openedx/frontend-component-header#663 is now making the notifications tray available out-of-the-box via a `HeaderNotificationsSlot`, removing the need for a separate bundled plugin in tutor-mfe.

### LLM usage notice


Faraz32123 (approved):
Changes look good! we can merge it.

PR #297, opened 2026-04-30
feat: add top-level COURSE_AUTHORING_MICROFRONTEND_URL LMS setting

dwong2708:
### Description

Set **COURSE_AUTHORING_MICROFRONTEND_URL** as a top-level Django setting in both the development and production **LMS** patches, in addition to MFE_CONFIG.

Previously it was only set inside MFE_CONFIG. This change mirrors the pattern already used for **ADMIN_CONSOLE_MICROFRONTEND_URL** and is required to support [openedx/openedx-platform#38415](https://github.com/openedx/openedx-platform/pull/38415), which reads **COURSE_AUTHORING_MICROFRONTEND_URL** directly from Django settings when building the studio_grading_url in the Instructor API v2.

### Changes

- openedx-lms-development-settings: define **COURSE_AUTHORING_MICROFRONTEND_URL** as a top-level setting and reference it in **MFE_CONFIG**
- openedx-lms-production-settings: same, with HTTPS-aware URL


arbrandes (approved):
👍🏼

PR #298, opened 2026-05-02
feat: use frontend-base compatibility layer

arbrandes:
### Description

Lets a frontend-base site keep working with FPF-built plugins as a migration aid during the MFE deprecation window. The shim is opt-in: plugin authors register an `(mfe_name, slot_name, plugin_config)` triple in `FRONTEND_COMPAT_SLOTS` to route a single slot through the compat layer, or list a Tutor plugin name in `FRONTEND_COMPAT_PLUGINS` to bulk-opt every contribution that plugin makes via `PLUGIN_SLOTS`. Sites with no opted-in contributions are unaffected.

Requires `tutor@main`: depends on the unreleased `ENV_SAVED` action.

### Implementation Details

Renders one `env.config.<mfe_name>.jsx` per MFE alongside the legacy per-MFE `env.config.jsx`, each carrying just the contributions opted in for that MFE. `site.config.{build,dev}.tsx` registers each one as its own shim `App` via `createLegacyPluginApp`, passing the MFE's name as `mfeId` so the shim can route-scope its contributions. The wildcard `"all"` is itself a valid `mfe_name`: it renders `env.config.all.jsx` and mounts it without an `mfeId`, so its contributions fire globally across every route, mirroring the legacy `iter_plugin_slots("all")` semantics. `site/package.json` adds the compat dep and an npm `overrides` block redirecting both `@openedx/frontend-plugin-framework` and `@edx/frontend-platform` to the same compat package, so transitive imports resolve to the shim's stubs and neither real package is installed.

Two new filters drive opt-in. `FRONTEND_COMPAT_SLOTS` is the per-slot opt-in registry, mirroring the `(mfe_name, slot_name, plugin_config)` shape of `PLUGIN_SLOTS`. `FRONTEND_COMPAT_PLUGINS` is a coarser bulk opt-in by Tutor plugin name, folding every `PLUGIN_SLOTS` contribution from that plugin's hook context through the shim.

Three more filters layer `(legacy_id, mapping)` deltas onto the shim's curated [`defaultSlotMap`](https://github.com/openedx/frontend-base-compat/blob/main/src/mappings/slotMap.ts), [`defaultWidgetMap`](https://github.com/openedx/frontend-base-compat/blob/main/src/mappings/widgetMap.ts), and [`defaultRouteMap`](https://github.com/openedx/frontend-base-compat/blob/main/src/mappings/routeMap.ts) for ids the shim doesn't yet cover: `FRONTEND_SLOT_COMPAT_MAPS`, `FRONTEND_WIDGET_COMPAT_MAPS`, and `FRONTEND_ROUTE_COMPAT_MAPS`. These deltas are site-wide (not per-MFE), so `site.config.{build,dev}.tsx` materializes the merged maps once and passes them by reference into every `createLegacyPluginApp` call. Identical contributions dedup silently; divergent ones warn and last-wins.

All compat wiring is gated on `get_frontend_compat_mfes()` being non-empty, and tagged `TODO(fpf-removal)` for cleanup on the FPF deprecation timeline. Full design captured in [openedx/frontend-base-compat ADR 0001](https://github.com/openedx/frontend-base-compat/blob/main/docs/decisions/0001-frontend-base-compatibility.rst).

### Testing

The easiest way to test this is with tutor-indigo, via this second PR:

https://github.com/overhangio/tutor-indigo/pull/219

After `tutor config save` and `tutor images build mfe`, you should see the indigoified header in the new Instructor Dashboard app:

image

### LLM usage notice


.github/workflows/test.yml
```diff
@@ -20,6 +20,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
+          pip install git+https://github.com/overhangio/tutor.git@main
```
arbrandes, line 23: We'll have to remember to remove this once it lands in Verawood.  But without it, CI will fail.

PR #309, opened 2026-06-03
improvement: optimized mfe production build

muhammadadeeltajamul:
Cache MFE build for production environment. Skip adding source maps in production build

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -763,6 +763,13 @@ Add any configurations at the end of the development webpack config file in Java
 
 File changed: ``tutormfe/templates/mfe/apps/mfe/webpack.dev-tutor.config.js``
 
+mfe-webpack-prod-config
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Add any configurations at the end of the production webpack config file in Javascript format.
+
+File changed: ``tutormfe/templates/mfe/build/mfe/webpack.prod-tutor.config.js``
+
 mfe-dockerfile-base
 ~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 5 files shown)


ahmed-arb:
Did you test this approach? Do we have any numbers on how this affects build time? Any caveats?

muhammadadeeltajamul:
> Did you test this approach? 

Yes, I created image builds after these changes. 

> Do we have any numbers on how this affects build time?

I don't have any numbers to show the improvement.

> Any caveats?

- This will require docker buildkit
- First build will have no major improvement as cache is not created initially. All subsequent builds will be faster as the cache will exist.

Abdul-Muqadim-Arbisoft:
Just added a change that made The webpack filesystem cache added for production builds was shared by every MFE. A BuildKit cache mount id defaults to the target path, and frontend-build sets no cache.name, so all 11 applications concurrently read and wrote a single "default-production" pack set (688 MB after a few builds).

The result was that the cache cost more than it saved: the first rebuild after a full image build was slower than having no cache at all.

Measured on authn, release/ulmo.3, 11 MFEs, Apple M1 Max:

no cache (baseline)   40.9s
shared cache          62.2s first rebuild / 15.3s steady state
per-MFE cache         17.1s first rebuild /  5.5s steady state
Replicated on profile and across two independent sessions: 3.0x faster on the first rebuild, 3.5x at steady state.

Also switch the mount to sharing=locked. Webpack's filesystem cache is not safe for concurrent writers, and with per-MFE ids the only remaining contention is two builds of the same MFE, which should serialise rather than race.

Abdul-Muqadim-Arbisoft (approved):

ahmed-arb (approved):

PR #327, opened 2026-08-26
feat: stop setting the ENABLE_MFE_CONFIG_API toggle

arbrandes:
### Description

Removes `ENABLE_MFE_CONFIG_API = True` from the LMS common settings patch. The toggle is deleted upstream in openedx/openedx-platform#39019, after which both `/api/mfe_config/v1` and `/api/frontend_site_config/v1/` respond unconditionally and setting the key does nothing.

This is a cleanup, not a lockstep requirement: upstream simply ignores the key once merged. But it must not merge before that PR does, since the toggle still defaults to `False` on master and dropping the line early would break MFE configuration. Hence the draft status.

`MFE_CONFIG_API_CACHE_TIMEOUT` is untouched and still applies.

### LLM usage notice


arbrandes:
This landed upstream, so we're good to go.

PR #333, opened 2026-08-31
fix: pin legacy MFE branches on master only

arbrandes:
### Description

The Authn, Learner Dashboard and Catalog legacy MFEs were pinned to their `legacy-mfe` branches unconditionally, which would break release builds, where those branches do not exist. The pin now only applies when `MFE_COMMON_VERSION` is `master`, resolved at render time by a new `get_mfe_version` template helper.

Plugins can pin the same way with a new optional `legacy_master` MFE attribute. An explicit `version` still wins in all cases.

### LLM usage notice

### DIFF
--- a/tutormfe/plugin.py
+++ b/tutormfe/plugin.py
@@ -73,7 +73,7 @@ CORE_MFE_APPS: dict[str, MFE_ATTRS_TYPE] = {
     "authn": {
         "repository": "https://github.com/openedx/frontend-app-authn.git",
         "port": 1999,
-        "version": "legacy-mfe",
+        "alternate_master": "legacy-mfe",
     },
     "authoring": {
         "repository": "https://github.com/openedx/frontend-app-authoring.git",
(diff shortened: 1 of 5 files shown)


tutormfe/plugin.py
```diff
@@ -430,6 +430,22 @@ def get_mfe(mfe_name: str) -> t.Union[MFE_ATTRS_TYPE, t.Any]:
...
+    """
+    Return the git revision an MFE should be built from.
+
+    An explicit "version" always wins. "alternate_master" only applies when
+    building against master: it names the branch to build from in place of
+    master, while release builds keep following the common version.
+    """
+    if "version" in mfe:
+        return mfe["version"]
+    if common_version == "master":
```
arbrandes, line 444: I don't _love_ this, but I also don't want to change more about the legacy MFEs than I have to.  The right way to do it when `__version_suffix__ == "main"` would be to not use MFE_COMMON_VERSION and let each checkout just get the default git branch unless a "main_version" is specified, but this would change behavior, too much.

Let's just use this hack for now until MFEs are removed.

PR #348, opened 2026-09-14
build: Dependency range must start with dev0

kdmccormick:
These version ranges were recently changed to
account for the fact that tutor now pre-releases
dev versions off of main. But, in order to to capture
the dev versions, the range must start with dev0.
This is because version sorting works like this:

X.Y.Z.devN < X.Y.Z.aN < X.Y.Z.bN < X.Y.Z.rcN < X.Y.Z

This fixes a new issue we're seeing on our sandboxes:
```
uv pip install git+https://github.com/overhangio/tutor-mfe.git@main
    Updated https://github.com/overhangio/tutor-mfe.git (e3b1a6a7635fc901f29dd3f11391198309
  × No solution found when resolving dependencies:
  ╰─▶ Because only tutor<=23.0.0.dev3180 is available and tutor-mfe==22.1.0 depends on
      tutor>=23.0.0,<24.0.0, we can conclude that tutor-mfe==22.1.0 cannot be used.
      And because only tutor-mfe==22.1.0 is available and you require tutor-mfe, we can
      conclude that your requirements are unsatisfiable.
```

This same fix may need to be applied to other places as well.


.github/workflows/test.yml
```diff
@@ -22,11 +22,6 @@ jobs:
         env:
           TARGET_BRANCH: ${{ github.base_ref || github.ref_name }}
         run: |
-          # On main this plugin requires the upcoming Tutor major version, which has
-          # no PyPI release yet: install Tutor from its main branch first.
-          if [ "$TARGET_BRANCH" = "main" ]; then
-            pip install "tutor[dev] @ git+https://github.com/overhangio/tutor@main"
-          fi
```
kdmccormick, line 29: I think we should remove this, because it's doing something in tests that we don't expect developers to have to do.

ahmed-arb (approved):
Thank you for the fix, and sorry for the churn. The range I landed was satisfiable only because everything that installed these branches had already put an unreleased Tutor in place, which is exactly the assumption a plain `pip install git+...@main` does not make.

.github/workflows/test.yml
```diff
@@ -22,11 +22,6 @@ jobs:
         env:
           TARGET_BRANCH: ${{ github.base_ref || github.ref_name }}
```
ahmed-arb, line 23: We can remove this as well, then.

kdmccormick:
Thanks @ahmed-arb ! I can't merge myself as I don't have write access on tutor-mfe

arbrandes (approved):

PR #350, opened 2026-09-19
fix: pin admin-console's legacy MFE branch on master only

arbrandes:
### Description

frontend-app-admin-console's `frontend-base` branch has taken over `master` per OEP-65 ADR 0004 (openedx/frontend-app-admin-console#226), and the micro-frontend now lives on `legacy-mfe`. Without this, nightly builds of the `admin-console` MFE stop working. Same shape as the Authn, Learner Dashboard, and Catalog entries; release builds keep following the common version.

### LLM usage notice


arbrandes:
Folded into #335, which ships the Admin Console as a frontend app now that its first alpha is on npm.

